PR #2, opened 2022-01-12
Add integration tests against a generated project

vierbergenlars:
To verify the correct workings of the starters, add an integration test that will build a standard generated content-cloud project.

There are *no* modifications done to the generated project, so it can easily be replaced with a freshly generated copy.

Instead, a build in the  `integration-tests` folder will grab the generated project and the main project and use them as included builds.
Using included builds automatically sets up dependency substitutions for the participating projects, and in addition we set up a configurable dependency substitution for the `org.springframework.boot` plugin itself, so we can test against different spring boot versions.


tgeens:
Integration tests fail because bintray (gradle plugins) is down again ?

vierbergenlars:
They *were* failing because bintray was down.

Now they are failing because the AWS SDK expects a region and throws when it does not have a region.
Is that something we can look to fix in the contentcloud-spring-boot-autoconfigure project? (which gets injected by the starter too)

tgeens:
Looks like it's a "missing feature" from the AWS SDK v2:

https://github.com/aws/aws-sdk-java-v2/issues/2229 which points to https://github.com/aws/aws-sdk-java-v2/issues/52
and we can conclude it's not getting fixed any time soon

So should we provide a _default_ region ?

vierbergenlars:
Since endpoint is configurable, we can put a bogus region in **if** the endpoint is set or S3 is disabled (until https://github.com/paulcwarren/spring-content/issues/667 is released).
Otherwise, we can read the region from a normal spring configuration variable and put it in the `aws.region` system property (that is read by s3)

contentcloud-spring-boot-autoconfigure/src/main/java/eu/xenit/contentcloud/spring/autoconfigure/s3/S3RegionAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+    public static class S3AdditionalProperties {
+        private String region;
+    }
+
+    /**
+     * This component will be registered (and created) before S3Client,
+     * and is thus able to set a property that is later picked up by the S3Client.
+     */
+    @Component
+    @AllArgsConstructor(onConstructor_= @Autowired)
```
tgeens, line 55: do you actually need that `@Autowired` annotation ?

contentcloud-spring-boot-autoconfigure/src/main/java/eu/xenit/contentcloud/spring/autoconfigure/s3/S3RegionAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+    static class S3RegionConfigurer implements InitializingBean {
+        private final S3Properties s3Properties;
+        private final S3AdditionalProperties s3AdditionalProperties;
+
+        @Override
+        public void afterPropertiesSet() throws Exception {
+            if(StringUtils.hasText(s3Properties.endpoint)) {
+                System.setProperty("aws.region", "fake");
```
tgeens, line 63: maybe `"none"` is more appropriate ?

contentcloud-spring-boot-autoconfigure/src/test/java/eu/xenit/contentcloud/spring/autoconfigure/s3/S3RegionAutoConfigurationTest.java
```diff
@@ -0,0 +1,118 @@
...
+        sdkClientConfigField.setAccessible(true);
+        return (SdkClientConfiguration) sdkClientConfigField.get(s3Client);
+    }
+
+    @Test
+    @ResourceLock(value = SYSTEM_PROPERTIES)
+    void setsRegionWhenConfiguredExplicitly() {
+        System.setProperty("spring.content.s3.region", "my-region");
+
+        AnnotationConfigApplicationContext applicationContext = new AnnotationConfigApplicationContext();
```
tgeens, line 56: FYI you could look into `ApplicationContextRunner` to test auto-configuration - that avoids that hassle with system-properties.

Something along the lines of:
```
new ApplicationContextRunner()
    .withConfiguration(AutoConfigurations.of(S3RegionAutoConfiguration.class))
    .withPropertyValues("spring.content.s3.region=my-region")
    .withUserConfiguration(TestConfig.class)
    .run(context -> {
        S3Client client = context.getBean(S3Client.class);
        SdkClientConfiguration configuration = extractConfiguration(client);
        assertEquals(Region.of("my-region"), configuration.option(AwsClientOption.AWS_REGION));
});
```

References:
* https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.developing-auto-configuration.testing
* https://www.baeldung.com/spring-boot-context-runner
* https://github.com/xenit-eu/thunx/pull/6/files#diff-72f43498992f08ffdb686bac323f0f964b31182a9ae330e8a8a543c463539a29

tgeens (approved):

contentcloud-spring-boot-autoconfigure/src/main/java/eu/xenit/contentcloud/spring/autoconfigure/s3/S3RegionAutoConfiguration.java
vierbergenlars: It looks like I don't need it

contentcloud-spring-boot-autoconfigure/src/test/java/eu/xenit/contentcloud/spring/autoconfigure/s3/S3RegionAutoConfigurationTest.java
vierbergenlars: Thanks for the references. I actually took a look at spring-content to see how to test autoconfiguration, but this is a lot better.

tgeens (approved):

PR #14, opened 2022-09-23
Publish with groupId com.contentgrid.starter

tgeens:


gradle/publish.gradle
```diff
@@ -30,25 +30,6 @@ publishing {
...
-                }
-            }
-        } else {
-            sonatypeSnapshots {
-                credentials {
-                    username = project.findProperty('sonatype_username')
-                    password = project.findProperty('sonatype_password')
-                }
-            }
-        }
```
rschev, line 50: I'm guessing the new plugin does this automatically?

rschev (approved):

gradle/publish.gradle
tgeens: It's in the main build.gradle now

PR #20, opened 2022-10-10
Update thunx dependency and gradle wrapper

rschev:


tgeens (approved):

vierbergenlars (changes requested):
I think you forgot to run `./gradlew wrapper` to actually update the `gradle-wrapper.jar` and the startup script.

vierbergenlars (approved):

PR #21, opened 2022-10-11
Custom actuators for policy rego file and webhook config

rschev:


contentgrid-spring-boot-actuators/src/main/java/com/contentgrid/spring/boot/actuator/policy/RegoProducible.java
```diff
@@ -0,0 +1,19 @@
+package com.contentgrid.spring.boot.actuator.policy;
+
+import org.springframework.boot.actuate.endpoint.Producible;
+import org.springframework.util.MimeType;
+
+public enum RegoProducible implements Producible<RegoProducible> {
```
vierbergenlars, line 6: There is no need to have this as an enum, just a class would be okay?

contentgrid-spring-boot-autoconfigure/build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
     annotationProcessor 'org.projectlombok:lombok'
 
     implementation platform(project(':contentgrid-spring-boot-platform'))
+    implementation project(':contentgrid-spring-boot-actuators')
```
vierbergenlars, line 25: I believe this would be a `compileOnly` dependency, with `ActuatorAutoConfiguration` gated behind `@ConditionalOnClass`

rschev:
If we get a third file that's _also_ structured like PolicyActuator.java and WebhooksConfigActuator.java, it'll be time to look at abstracting the common code away a little more. It didn't seem necessary to me (yet) with only two classes.

contentgrid-spring-boot-platform/build.gradle
```diff
@@ -13,7 +13,9 @@ ext {
 
 dependencies {
     api platform("org.springframework.boot:spring-boot-dependencies:2.6.12")
-    api platform("com.contentgrid.thunx:thunx-dependencies:0.4.0")
+    api platform("com.contentgrid.thunx:thunx-bom:0.4.2")
+
+    api "org.springframework.boot:spring-boot-actuator"
```
vierbergenlars, line 18: This is a platform, direct dependencies should only be other platforms, not actual dependencies.

The rest of version constraints should be under `constraints`.

In this case, the `spring-boot-actuator` version is already set by the `spring-boot-dependencies` platform, so it is not necessary at all.

contentgrid-spring-boot-actuators/src/main/java/com/contentgrid/spring/boot/actuator/policy/RegoProducible.java
rschev: If I understand `interface Producible<E extends Enum<E> & Producible<E>>` correctly, it needs to be an enum.

From [org.springframework.boot.actuate.endpoint.Producible](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Producible.java)

contentgrid-spring-boot-autoconfigure/build.gradle
tgeens: yes - typically the `-autoconfigure` project has no runtime dependencies, it's the `-starter` that pulls in the (runtime) dependencies

contentgrid-spring-boot-actuators/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+}
+
+dependencies {
+    annotationProcessor platform(project(':contentgrid-spring-boot-platform'))
+    annotationProcessor 'org.projectlombok:lombok'
+
+    implementation platform(project(':contentgrid-spring-boot-platform'))
+    implementation 'org.springframework.boot:spring-boot-autoconfigure'
+
+    api 'org.springframework.boot:spring-boot-actuator'
```
tgeens, line 27: `implementation` dependency ?

tgeens: or `compileOnly` ?

tgeens: we're exposing `Producible<>` from `spring-boot-actuator` here, so this needs to be `api`

tgeens (approved):

PR #23, opened 2022-11-02
Fix transitive dependency on actuators

rschev:
Before this, consumers needed to explicitly depend on contentgrid-spring-boot-actuators as well. Now they can just depend on the starter again.

### DIFF
--- a/contentgrid-spring-boot-platform/build.gradle
+++ b/contentgrid-spring-boot-platform/build.gradle
@@ -18,6 +18,7 @@ dependencies {
     constraints {
         api project(':contentgrid-spring-boot-starter')
         api project(':contentgrid-spring-boot-starter-annotations')
+        api project(':contentgrid-spring-boot-actuators')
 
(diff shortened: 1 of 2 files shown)


tgeens (approved):

contentgrid-spring-boot-actuators/build.gradle
```diff
@@ -35,3 +35,12 @@ dependencies {
 tasks.named('test') {
     useJUnitPlatform()
 }
+publishing {
+    publications.all {
+        versionMapping {
+            allVariants {
+                fromResolutionResult()
```
vierbergenlars, line 42: Are you sure that this is necessary?

The other projects were also published without `fromResolutionResult()`, and they worked, right?

rschev: I got it from the annotations build.gradle, but it indeed turns out to be unnecessary for this.

PR #24, opened 2022-11-02
integration events starter

dgradecak:


contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+        return new ContentGridPublisherEventListener(contentGridEventPublisher,
+                entityManagerFactory);
+    }
+
+    @Configuration
+    @ConditionalOnProperty(value = { "spring.rabbitmq.host" })
+    static class EventsRabbitMqAutoConfiguration {
+
+        @Bean
+        @ConditionalOnProperty(value = { "contentgrid.events.queueName" })
```
vierbergenlars, line 72: Why is the creation of this Queue been dependent on `contentgrid.events.queueName` being set? Isn't it set to a default value by default?

dgradecak: yes there is a default value, but I think it might be good to have the smae approach for mq host. So, if it is not set in the config, than we do not create a queue instance and expect the default queue to be present. Not sure is it makes sense though ...

vierbergenlars: Oh, this only creates a queue on the activemq side? It is not necessary to be able to send messages?

tgeens: First: it is rabbitmq, not activemq ;-)

If I understand the rabbitmq model correctly:
- a publisher sends to a (named) _exchange_ 
- rabbitmq is configured to route messages from an exchange to one or more queues
- consumers read from the queue

Then why are we setting up the queue on the producer-side ?

dgradecak: > Oh, this only creates a queue on the activemq side? It is not necessary to be able to send messages?

yes, if the queue is not created than the message cannot be sent

dgradecak: > Then why are we setting up the queue on the producer-side ?

same reason as above;) The thing is if the queue is not crated (automatically by this config, or manually) the message will not be able to be delivered. I am not aware of any other mechanism that we can use beside this one, unless rabbitmq can create a queue on the fly somehow, maybe there is some config

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventConfigurationProperties.java
```diff
@@ -0,0 +1,19 @@
+package com.contentgrid.spring.boot.autoconfigure.integration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("contentgrid.events")
+public class EventConfigurationProperties {
+
+    static public final String DEFAULT_QUEUE_NAME = "content-grid.events";
```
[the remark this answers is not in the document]
dgradecak: hehe I should know that by now

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
tgeens: > The thing is if the queue is not created (automatically by this config, or manually) the message will not be able to be delivered. 

True. But I think the queue should be created by the webhooks-server (if, and only if it is present ?)

tgeens:
:point_up_2: rebased on top of `master`

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+        Queue queue(EventConfigurationProperties config) {
+            return new Queue(config.getQueueName(), false);
+        }
+
+        @Bean
+        // TODO do we need a conditional here?
+        ContentGridMessageHandler messageHandler(RabbitTemplate rabbitTemplate,
+                EventConfigurationProperties config) {
+            return () -> Amqp.outboundAdapter(rabbitTemplate).routingKey(config.getQueueName());
```
tgeens, line 88: so technically `EventConfigurationProperties` should not have a `queue-name` property, but a `routing-key` property instead ?

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
tgeens: I want to have the message land in rabbitmq exchange - that is the end of the contract.
If there is no route/queue configured, rabbitmq can blackhole it  :shrug:

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
tgeens: fair point - but can we use `contentgrid.events.queue-name` instead :-)

tgeens: `contentgrid.events.rabbitmq.routing-key`

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
dgradecak: as agreed, the starter is not responsible creating the queue. The queue should be there and exposed by the webhooks app

tgeens:
Can I still start a boot-app with this starter without any rabbitmq ?
Like, I should set some property to enable rabbitmq ?

dgradecak:
> Can I still start a boot-app with this starter without any rabbitmq ? Like, I should set some property to enable rabbitmq ?

yes, you first need to add a dependency to "com.contentgrid.starter:contentgrid-spring-integration-events" and second, you have to add set  "spring.rabbitmq.host"

tgeens (approved):

PR #28, opened 2022-11-14
Include sources and javadoc, add scm.connection

rschev:
From my test, it _seems_ that this is the last one before it truly is fixed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,14 @@ allprojects {
     }
     pluginManager.withPlugin('java') {
         java {
-            sourceCompatibility = JavaVersion.VERSION_11
+            sourceCompatibility = JavaVersion.VERSION_17
+        }
+    }
+    pluginManager.withPlugin('java-library') {
+        java {
+            sourceCompatibility = JavaVersion.VERSION_17
+            withSourcesJar()
+            withJavadocJar()
         }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -11,14 +11,11 @@ allprojects {
     pluginManager.withPlugin('maven-publish') {
         apply from: "${rootDir}/gradle/publish.gradle"
     }
-    pluginManager.withPlugin('java') {
-        java {
-            sourceCompatibility = JavaVersion.VERSION_17
-        }
-    }
     pluginManager.withPlugin('java-library') {
         java {
-            sourceCompatibility = JavaVersion.VERSION_17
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -16,4 +16,32 @@ allprojects {
             sourceCompatibility = JavaVersion.VERSION_11
         }
     }
+    pluginManager.withPlugin('java-library') {
+        configurations {
+            // See https://github.com/gradle/gradle/issues/10861
+            // To avoid publishing the internal-platform
+            internalPlatform {
```
tgeens, line 23: now you are just copy/pasting gradle snippets from other projects ? :joy:

Why are you defining `internalPlatform` here, when it is not used anywhere in this project ?

build.gradle
```diff
@@ -13,7 +13,28 @@ allprojects {
...
+            sourceCompatibility = JavaVersion.VERSION_17
+        }
+    }
+    pluginManager.withPlugin('java-library') {
+        configurations {
+            compileClasspath.extendsFrom(internalPlatform)
+            runtimeClasspath.extendsFrom(internalPlatform)
+            testCompileClasspath.extendsFrom(internalPlatform)
+            testRuntimeClasspath.extendsFrom(internalPlatform)
+        }
```
tgeens, line 25: more `internalPlatform` ?

build.gradle
```diff
@@ -13,7 +13,21 @@ allprojects {
...
+        java {
+            sourceCompatibility = JavaVersion.VERSION_17
+            withSourcesJar()
+            withJavadocJar()
+        }
+
+        repositories {
+            mavenCentral()
+            maven {
+                url 'https://s01.oss.sonatype.org/content/repositories/snapshots/'
```
tgeens, line 29: why do we have this ?

build.gradle
```diff
@@ -13,7 +13,19 @@ allprojects {
     }
     pluginManager.withPlugin('java') {
         java {
-            sourceCompatibility = JavaVersion.VERSION_11
+            toolchain {
+                languageVersion = JavaLanguageVersion.of(17)
+            }
+        }
+
+    }
```
tgeens, line 21: you can remove this one - it's not used in this build
(it is used in `integration-tests-spring` but that is actually a separate build)

tgeens (approved):
one more thing to cleanup and then you can merge ;-)

PR #30, opened 2022-11-18
Add support for separating user-provided template variables from system variables

rschev:


contentgrid-spring-boot-actuators/src/main/java/com/contentgrid/spring/boot/actuator/TemplateHelper.java
```diff
@@ -0,0 +1,36 @@
+package com.contentgrid.spring.boot.actuator;
+
+import com.contentgrid.spring.boot.actuator.policy.PolicyTemplatingProperties;
+import com.contentgrid.spring.boot.actuator.webhooks.WebhooksTemplatingProperties;
+import lombok.RequiredArgsConstructor;
+import org.springframework.util.PropertyPlaceholderHelper;
+
+@RequiredArgsConstructor
+public class TemplateHelper {
```
tgeens, line 9: I'm not a fan of coupling both the webhook system en the policies system like this

tgeens: if you need this helper, can this please be generic ?

rschev: Would you prefer
1. Two TemplateHelper beans, distinguished via @Qualifier, one for policy and another for webhooks
2. One TemplateHelper bean, which only has the system TemplatingProperties as a member variable, and any other TemplatingProperties you want need to be provided as a parameter to the `template(…)` method

tgeens (approved):
conditional approve on renaming `SystemProperties` to something like `ContentGridApplicationProperties`

contentgrid-spring-boot-actuators/src/main/java/com/contentgrid/spring/boot/actuator/SystemProperties.java
```diff
@@ -0,0 +1,11 @@
+package com.contentgrid.spring.boot.actuator;
+
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class SystemProperties {
```
vierbergenlars: This is a confusing name, because "system" properties also means the `system.deployment.id` properties etc.

PR #34, opened 2022-12-07
[ACC-449] Re-organize policy & webhook properties

vierbergenlars:
Split between configuration properties from spring-boot and variables used by policy & webhook actuators


contentgrid-spring-boot-actuators/src/main/java/com/contentgrid/spring/boot/actuator/webhooks/WebhookVariables.java
```diff
@@ -0,0 +1,29 @@
...
+        if(placeholderName.startsWith("vars.")) {
+            return userVariables.get(placeholderName.substring("vars.".length()));
+        }
+        switch(placeholderName) {
+            case "system.application.id":
+                return systemProperties.getApplicationId();
+            case "system.deployment.id":
+                return systemProperties.getDeploymentId();
+            default:
+                throw new IllegalArgumentException(String.format("Can not find a replacement for placeholder '%s'", placeholderName));
```
tgeens, line 26: that kind of violates the `PlaceholderResolver` contract. I'm not sure this is the correct place to throw an error when some placeholder cannot be resolved.

tgeens (approved):

PR #78, opened 2023-06-02
Disable Spring Content shortcut request mapping

tgeens:


contentgrid-spring-data-rest/build.gradle
```diff
@@ -35,21 +35,30 @@ dependencies {
...
+    testFixturesApi 'com.github.paulcwarren:spring-content-rest'
+
+
     testFixturesImplementation 'org.springframework.boot:spring-boot-starter-data-jpa'
     testFixturesImplementation 'org.springframework.boot:spring-boot-starter-data-rest'
-    testFixturesImplementation 'org.projectlombok:lombok'
+    testFixturesImplementation "com.github.paulcwarren:spring-content-rest-boot-starter"
+    testFixturesImplementation "com.github.paulcwarren:spring-content-fs-boot-starter"
 
+    testFixturesImplementation 'org.projectlombok:lombok'
```
vierbergenlars, line 61: Lombok is a compile-time only annotation processor, so it should not be pulled in as an implementation dependency

contentgrid-spring-data-rest/build.gradle
```diff
@@ -35,21 +35,30 @@ dependencies {
...
 
+
+
     testFixturesAnnotationProcessor 'com.querydsl:querydsl-apt::jpa'
     testFixturesAnnotationProcessor 'jakarta.persistence:jakarta.persistence-api'
 
     testFixturesImplementation platform(project(':contentgrid-spring-boot-platform'))
 
     testFixturesApi 'org.springframework.data:spring-data-jpa'
+    testFixturesApi 'com.github.paulcwarren:spring-content-commons'
```
vierbergenlars, line 52: These are only used for annotations on the model and repositories.

I think they don't have to be an API dependency then, I think annotations are not part of the exposed API.

vierbergenlars (approved):

PR #86, opened 2023-07-04
Events for collections [ACC-536]

rschev:
Unfortunately, these events say very little and do not contain a
meaningful before/after or diff. There appears to be no way for us to
get this data without reading out the entire join table and the one from
the snapshot, and diffing on that. This is a very expensive operation in
case of a large table.


vierbergenlars (commented):
Needs a unit or integration test

integration-tests/integration-tests-spring/src/test/java/com/contentgrid/userapps/holmes/dcm/repository/DcmApiRepositoryIntegrationEventsTests.java
```diff
@@ -179,6 +188,24 @@ void whenCaseIsDeleted_postUpdateShouldBeCalledOnce_ok() {
         verify(listener, times(1)).onPostInsert(any());
         verify(listener, times(1)).onPostUpdate(any());
         verify(listener, times(1)).onPostDelete(any());
+        verify(listener, times(0)).onPostUpdateCollection(any());
+
+        verify(contentGridMessageHandler.get().get(), times(3)).handleMessage(any());
+    }
+
+    @Test
+    void whenPersonIsAddedToCaseSuspects_postUpdateShouldBeCalledOnce_ok() {
```
vierbergenlars, line 197: Can you also add tests for one-to-many & many-to-one relations, that those don't trigger multiple events?

vierbergenlars (approved):

PR #92, opened 2023-07-25
Fall back to spring-data-rest implementation for non-implemented cases in DelegatingRepositoryPropertyReferenceController

vierbergenlars:
This is necessary to bridge the gap while bi-directional relations are not yet fully supported.
It default to disabled, because it also has security implications when used in conjunction with thunx

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
+++ b/contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
@@ -69,11 +69,14 @@ public class DelegatingRepositoryPropertyReferenceController {
 
     private final ObjectProvider<QuerydslBindingsFactory> querydslBindingsFactoryProvider;
 
+    private final ContentGridRestProperties contentGridRestProperties;
+
     @Autowired
     DelegatingRepositoryPropertyReferenceController(RepositoryPropertyReferenceController delegate,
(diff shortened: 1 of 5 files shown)


contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
```diff
@@ -283,6 +294,16 @@ public ResponseEntity<Void> handle(
...
 
+    private ResponseEntity<?> fallbackIfRequired(ResponseEntity<?> original, ThrowingSupplier<ResponseEntity<?>> defaultSupplier) throws Exception {
+        if(!contentGridRestProperties.isFallbackToDefaultRelationController()) {
+            return original;
+        }
+        if(original.getStatusCode() == HttpStatus.NOT_IMPLEMENTED) {
+            return defaultSupplier.get();
+        }
+        return original;
+    }
```
rschev, line 305: I feel like the wording could be improved, _default_ and _original_ don't really contrast each other in an obvious way.
What do you think of
```
fallbackIfRequired(ResponseEntity<?> cgResponse, ThrowingSupplier<ResponseEntity<?>> springDRDefaultSupplier) { ... }
```
?

Additionally, I think _fallbackIfRequired_ could be _fallbackIfNotImplemented_ or something like that? "If Required" sounds like it will always pick the second option if the config property is set.

rschev (approved):

PR #98, opened 2023-08-01
Add custom CurieProvider

vierbergenlars:
- Add custom curie provider that does not automatically curie links
- Add utility-methods to CurieProviderCustomizer to register a curie
- Make CurieProvider validate registered curie prefixes and link relations
- Add javadoc to CurieProvider{Builder,Customizer} interfaces

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/hal/ContentGridCurieProvider.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/hal/ContentGridCurieProvider.java
@@ -4,6 +4,8 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import lombok.RequiredArgsConstructor;
+import org.springframework.hateoas.IanaLinkRelations;
+import org.springframework.hateoas.IanaUriSchemes;
 import org.springframework.hateoas.Link;
 import org.springframework.hateoas.LinkRelation;
 import org.springframework.hateoas.Links;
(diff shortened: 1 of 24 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/hal/ContentGridCurieProvider.java
```diff
@@ -0,0 +1,95 @@
...
+        curies.put(prefix, template);
+        return new ContentGridCurieProvider(curies);
+    }
+
+    @Override
+    public CurieProvider build() {
+        return this;
+    }
+
+    private void checkRegisteredCurie(LinkRelation rel) {
```
rschev, line 72: I kind of expect a method with `check…` to return a boolean. Maybe `assertRegisteredCurie` or `throwIfNotRegisteredCurie` instead?

contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/Order.java
```diff
@@ -34,19 +34,23 @@ public class Order {
     @ManyToOne
     @JoinColumn(name = "customer")
     @CollectionFilterParam
+    @org.springframework.data.rest.core.annotation.RestResource(rel = "d:customer")
```
rschev, line 37: Why fully qualified instead of importing?

vierbergenlars: Well, because there is a second one also named `RestResource` (from spring-content) that is imported.

rschev (approved):

PR #99, opened 2023-08-02
Add support for generic link relations based on datamodel

vierbergenlars:
These generic link relations: `cg:entity`, `cg:relation` & `cg:content` can be used to build generic integrations that are independent of the datamodel of the application.

Such a generic application can now easily read all content properties from an entity without relying on heuristics to determine which link-relations on an entity object are content and which ones are other relations.

Similarly, a generic application can now discover all available entity types from `/` (or their profiles from `/profile`) by using the `cg:entity` relation, instead of reading all relations and assuming which ones are entities based on heuristics.

Since the generic link relations rely on CURIEs so they are namespaced, they are only activated when a `CurieProviderCustomizer` is defined (and thus when all custom links based on the datamodel are being prefixed)

### DIFF
--- a/contentgrid-spring-boot-autoconfigure/build.gradle
+++ b/contentgrid-spring-boot-autoconfigure/build.gradle
@@ -48,6 +48,7 @@ dependencies {
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
     testImplementation 'org.springframework.data:spring-data-rest-webmvc'
     testImplementation 'com.github.paulcwarren:spring-content-s3-boot-starter'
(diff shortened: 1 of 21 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/links/ContentGridSpringDataLinksConfiguration.java.orig
rschev, line 1: :thinking:

rschev (approved):

PR #100, opened 2023-08-11
Use 'item' link-rel for collection `_embedded` items

vierbergenlars:
- Use 'item' as link-relation for collection resources
- Add 'Refund' model to fixture application
- Ensure that links are present on the copied model


tgeens (approved):

tgeens:
Ref https://www.rfc-editor.org/rfc/rfc6573.html

PR #101, opened 2023-08-16
Upgrade to Spring Boot 3

vierbergenlars:
- Upgrade to spring boot 3
- Disable check for content-disposition of spring-content response


contentgrid-spring-boot-platform/build.gradle
```diff
@@ -9,9 +9,9 @@ javaPlatform {
 
 
 dependencies {
-    api platform('org.springframework.boot:spring-boot-dependencies:2.7.14')
+    api platform('org.springframework.boot:spring-boot-dependencies:3.0.7')
```
tgeens, line 12:
```suggestion
    api platform('org.springframework.boot:spring-boot-dependencies:3.0.9')
```

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/test/fixture/invoicing/InvoicingApplicationTests.java
```diff
@@ -888,8 +888,11 @@ void getInvoiceContent() throws Exception {
                         .andExpect(status().isOk())
                         .andExpect(content().contentType(MIMETYPE_PLAINTEXT_UTF8))
                         .andExpect(content().string(EXT_ASCII_TEXT))
+                        ;
+                        /* This assertion is changed in SB3; and is technically incorrect
+                        (it should be `Content-Disposition: attachment` or `Content-Disposition: inline` with a filename, never `form-data`)
                         .andExpect(headers().string("Content-Disposition",
-                                is("form-data; name=\"attachment\"; filename*=UTF-8''%s".formatted(encodedFilename))));
```
tgeens, line 892: so the assertion was correct before, but a regression in spring-content ?

tgeens (approved):

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/test/fixture/invoicing/InvoicingApplicationTests.java
vierbergenlars: This is not a regression in spring-content. The assertion itself is now failing because spring adds a `filename=[...]` part as well.

However, the semantics of the assertion are incorrect: 

`content-disposition` set to `form data` is only valid in requests as a form field, never in a response: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_header_for_a_multipart_body

In a response, `content-disposition` should be set to either `inline` (show in browser) or `attachment` (perform download). In both cases, the `filename` (and `filename*`) can be set as well.

I am going to send a PR to spring-content to fix this, but first wanted to get this upgrade out of the way

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/integration/EventsAutoConfiguration.java
```diff
@@ -50,7 +50,7 @@ IntegrationFlow contentGridEventsFlow(ObjectProvider<ContentGridMessageHandler>
 
         ObjectMapper halObjectMapper = typeConstrainedMappingJackson2HttpMessageConverter.getObjectMapper();
 
-        IntegrationFlowBuilder builder = IntegrationFlows
+        IntegrationFlowBuilder builder = IntegrationFlow
```
tgeens, line 53: weird change

tgeens (approved):

PR #105, opened 2023-08-17
Configure flyway to disable transactional lock

vierbergenlars:
There are some changes made in Flyway between the version shipped in Spring Boot 2 & 3
A default was changed in flyway which results in `CREATE INDEX CONCURRENTLY` in a migration deadlocking during startup

The resolution provided in https://github.com/flyway/flyway/issues/3508 is to set a configuration option for transactional lock to false.


vierbergenlars:
It's a trade-off made by flyway. Transactional locking is a *new feature* introduced by flyway.

Disabling transactional locks does not disable transactionality of migrations, nor does it disable locking. It is essentially a switch between a transactional lock and a session-based lock: https://documentation.red-gate.com/fd/postgresql-transactional-lock-184127530.html

In essence, the new transactional locking feature was created to support connecting through PGBouncer that does connection-pooling: https://github.com/flyway/flyway/issues/2895

tgeens (approved):

PR #112, opened 2023-08-29
Replace spring data QuerydslBinder with CollectionFilter-based search filters

vierbergenlars:
Using the CollectionFilter-based filters allow a couple of additional features:
 - Only having to define the filters in one place (no more `QuerydslBinderCustomizer` in repositories)
 - Support multiple query parameters bound to the same `Path`, with different actions (like being able to do greater than/less than queries)


contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/ContentGridSpringDataRestConfiguration.java
```diff
@@ -52,6 +55,22 @@ public Object postProcessAfterInitialization(Object bean, String beanName) throw
         };
     }
 
+    @Bean
+    public BeanPostProcessor replaceQuerydslWithCollectionFilterParamPredicateResolver(ApplicationContext applicationContext) {
```
tgeens, line 59: Is this temporary?  It looks pretty wrong if you have to resolve to BeanPostProcessor-tricks for your own beans :grimacing:

tgeens (approved):

PR #116, opened 2023-08-30
Call correct delegate method in `AggregateLinkCollector#getLinksForNested()`

vierbergenlars:
Calling the `getLinksFor()` delegate method was not intended, and results in a problem when having `@Embedded` objects, as those have no self link, nor an ID to put in a self link


tgeens (approved):

vierbergenlars:
It's covered by tests now (even though they are integration tests)

PR #117, opened 2023-08-31
Add case-insensitive predicate

vierbergenlars:
- Fixup types of CollectionFilterParam#predicate
- Add EqualsIgnoreCase predicate that does case-insensitive string equality comparisons

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/querydsl/mapping/BeanFactoryPredicateFactoryInstantiator.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/querydsl/mapping/BeanFactoryPredicateFactoryInstantiator.java
@@ -11,9 +11,8 @@ public class BeanFactoryPredicateFactoryInstantiator implements PredicateFactory
     private final PredicateFactoryInstantiator fallback;
 
     @Override
-    public QuerydslPredicateFactory<Path<?>, ?> instantiate(Class<? extends QuerydslPredicateFactory<Path<?>, ?>> clazz) {
-        Class c = clazz;
-        return (QuerydslPredicateFactory<Path<?>, ?>) beanFactory.getBeanProvider(c)
+    public QuerydslPredicateFactory<? extends Path<?>, ?> instantiate(Class<? extends QuerydslPredicateFactory> clazz) {
+        return (QuerydslPredicateFactory<? extends Path<?>, ?>) beanFactory.getBeanProvider((Class)clazz)
                 .getIfAvailable(() -> fallback.instantiate(clazz));
     }
 }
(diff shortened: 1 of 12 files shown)


contentgrid-spring-querydsl/src/main/java/com/contentgrid/spring/querydsl/predicate/AbstractSimpleQuerydslPredicateFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        return Stream.of(coercePath(path));
+    }
+
+    @Override
+    final public Optional<Predicate> bind(Path<?> path, Collection<? extends S> values) {
+        return bindCoerced(coercePath(path), values);
+    }
+
+    /**
+     * Converts the generic path type that {@link com.contentgrid.spring.querydsl.annotation.CollectionFilterParam} to a specific type that {@link #bindCoerced(Path, Collection)} can work upon
```
tgeens, line 33: I'm having trouble parsing this sentence

contentgrid-spring-querydsl/src/test/java/com/contentgrid/spring/querydsl/predicate/DefaultTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.springframework.data.convert.Jsr310Converters;
+
+class DefaultTest {
+    private final Default DEFAULT_FACTORY = new Default();
+
+    private final PredicateFactoryTester<QTestObject> TESTER = new PredicateFactoryTester<>(new QTestObject(PathMetadataFactory.forVariable("o")));
+
+    private static final ConfigurableConversionService conversionService = new DefaultConversionService();
+    static {
+        Jsr310Converters.getConvertersToRegister().forEach(conversionService::addConverter);
```
tgeens, line 32: why do we need this ? the javadoc says it is only needed if you run on java 8 ?

tgeens (approved):

contentgrid-spring-querydsl/src/test/java/com/contentgrid/spring/querydsl/predicate/DefaultTest.java
vierbergenlars: That javadoc is [from 9 years ago (2014)](https://github.com/spring-projects/spring-data-commons/blame/d0ba1252687acf416c838cef840b4145ab859814/src/main/java/org/springframework/data/convert/Jsr310Converters.java#L41), which is around the time that Java 8 was released. I guess it was used to avoid runtime class not found errors when running on Java 7 or earlier.

In fact, the javadoc was [updated recently](https://github.com/spring-projects/spring-data-commons/commit/dc999a81937bf05f2624fb5919e36b77027f3c22) to remove the Java 8 reference.

contentgrid-spring-querydsl/src/main/java/com/contentgrid/spring/querydsl/predicate/AbstractSimpleQuerydslPredicateFactory.java
vierbergenlars: I think I got interrupted while writing that sentence.

Is this better?

```suggestion
     * Converts the generic path type to a specific type that {@link #bindCoerced(Path, Collection)} can work upon
```

PR #118, opened 2023-09-01
Support one-to-many redirect with pseudo-inverse relation using a plain Column

vierbergenlars:
To support permissions on -to-many relations, we have a feature in to automatically redirect to collection of the target of the relation, with a collection filter applied.
For example: `/invoices/$id/orders` becomes `/orders?invoice._id=$id`

To be able to apply a collection filter on the target domain type, that type needs a reference to the source domain type. (so, a bi-directional relation)
We don’t want to implement full user-configurable bi-directional relations for now, since that requires additional considerations.

The complications of bi-directional relations in JPA:
- The inverse side of a relation is read-only: Hibernate does not look at the inverse side of the relation when saving changes
- The `@ManyToOne` relation annotation does not have a `mappedBy` attribute, so in a bi-directional relation it is always the primary side; `@OneToMany` will always be the inverse side in a bi-directional relation. Uni-directional `@OneToMany` is fine and there it can be the primary side.

If we limit the scope of bi-directional relations to the ones required to support permissions on relations:
- `@ManyToOne`: No problem, we redirect directly to the “one” item on the target side, we don’t need filters
- `@OneToOne`: No problem, identical to `@ManyToOne`
- `@ManyToMany`: No problem, the side on the target entity can be the inverse side. We mark it as not exported, so it’s not visible in the API response, and it’s not writable
- `@OneToMany`: Ouch, according to our model we need a `@ManyToOne` on the inverse side, but from the perspective of JPA, that needs to be the primary side. That would mean that API users can’t write to the relation anymore from the primary side of the model (and the other side is not visible in the API)

To solve this problem, we need to keep the `@OneToMany` side the JPA primary side. Technically, the data is stored in the table of the target. We point there with a `@JoinColumn` annotation.

Since technically the data is already present on the table of the target entity, we should be able to access it from that side as well, but we can’t make it a `@ManyToOne` relation. Since it is stored in the table of the entity, a plain `@Column` annotation enables us to pull the data in.
It is a bit of a wonky setup, but if we take care to only read data from there, and don’t rely on it being updated within the same JPA session, it should be fine.
Since writing and reading data always happens from separate HTTP requests, we can be sure that they are separate sessions and we don’t need to worry about this.

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
+++ b/contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
@@ -9,6 +9,8 @@ import static org.springframework.web.bind.annotation.RequestMethod.POST;
 import static org.springframework.web.bind.annotation.RequestMethod.PUT;
 
 import com.contentgrid.spring.querydsl.mapping.CollectionFiltersMapping;
+import jakarta.persistence.Column;
+import jakarta.persistence.JoinColumn;
 import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.net.URI;
@@ -97,7 +99,9 @@ public class DelegatingRepositoryPropertyReferenceController {
                     log.warn("Could not find other side of relation {}", prop.property);
                     return ResponseEntity.status(HttpStatus.NOT_IMPLEMENTED).build();
                 }
-                var idPropertyFilter = collectionFiltersMapping.forIdProperty(targetType.getType(), mappedBy.get());
+
+                var idPropertyFilter = collectionFiltersMapping.forProperty(targetType.getType(), mappedBy.get())
+                        .or(() -> collectionFiltersMapping.forIdProperty(targetType.getType(), mappedBy.get()));
 
                 if (idPropertyFilter.isPresent()) {
                     var filter = idPropertyFilter.get().getFilterName();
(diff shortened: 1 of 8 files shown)


contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
```diff
@@ -127,14 +131,27 @@ public ResponseEntity<?> followPropertyReference(final RootResourceInformation r
...
             }
 
+            // It's also possible there is an implicit relation with the other side being a plain @Column that matches our @JoinColumn
+            var joinColumn = property.findAnnotation(JoinColumn.class);
+            if(joinColumn != null) {
+                var ourColumnName = joinColumn.name();
+                var inverseSide = repositories.getPersistentEntity(property.getAssociationTargetType());
+                for (PersistentProperty<? extends PersistentProperty<?>> inverseProperty : inverseSide.getPersistentProperties(
+                        Column.class)) {
+                    if (Objects.equals(ourColumnName, inverseProperty.findAnnotation(Column.class).name())) {
```
rschev, line 149: `findAnnotation` returns an Optional, you need to destructure it before the equality check.

vierbergenlars: 1. `Property#findAnnotation()` returns the value or null.
2. Given we use `getPersistentProperties(Column.class)`, we only get properties that have a `@Column` annotation, so `#findAnnotation()` will never return null.

tgeens:
Do we have somewhere test coverage for when `@Column` is not present?

tgeens (approved):

contentgrid-spring-data-rest/src/main/java/org/springframework/data/rest/webmvc/DelegatingRepositoryPropertyReferenceController.java
rschev: Yeah it became clear in intellij. For some reason, github's popup code thingy when clicking the method assured me that it returns an Optional.

vierbergenlars:
> Do we have somewhere test coverage for when `@Column` is not present?

Not at this point. The current behavior should be to ignore fields where no `@Column` is present, even when the derived column name on the other side would match the `@JoinColumn` annotation

PR #119, opened 2023-09-01
Add conditional annotations to load less when running as initcontainer

rschev:


contentgrid-spring-boot-autoconfigure/build.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
 
     implementation platform(project(':contentgrid-spring-boot-platform'))
     implementation 'org.springframework.boot:spring-boot-autoconfigure'
+    implementation 'org.springframework.boot:spring-boot-actuator-autoconfigure'
```
vierbergenlars, line 31:
```suggestion
    compileOnly 'org.springframework.boot:spring-boot-actuator-autoconfigure'
```

We don't want to pull in actuators autoconfiguration when the user does not have actuators on the classpath.

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/actuator/ActuatorAutoConfiguration.java
```diff
@@ -28,6 +29,7 @@
 
 @AutoConfiguration
 @ConditionalOnClass({PolicyActuator.class, WebHooksConfigActuator.class})
```
vierbergenlars, line 31:
```suggestion
@ConditionalOnClass({PolicyActuator.class, WebHooksConfigActuator.class, ConditionalOnAvailableEndpoint.class})
```

And here, we're making sure this thing is not imported when spring-boot-actuator-autoconfigure is not on the classpath

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/actuator/ActuatorAutoConfiguration.java
```diff
@@ -28,6 +29,7 @@
...
+@ConditionalOnAvailableEndpoint(endpoint = PolicyActuator.class)
```
vierbergenlars, line 32: Doing this will also disable the webhooks actuator when only the policy actuator is disabled.

I don't think we want to be doing that?

Instead, you can create two inner classes with `@ConditionalOnAvailableEndpoint` (one for policy and one for webhook), and only the dependencies of the respective endpoints

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/data/web/ContentGridSpringDataRestAutoConfiguration.java
```diff
@@ -32,7 +34,8 @@
                 // is initialized so @ConditionalOnBean works correctly. Putting the annotation directly on that class
                 // does not work, because it is not an autoconfiguration, but is initialized directly when this parent class
                 // is initialized.
-                "internal.org.springframework.content.rest.boot.autoconfigure.ContentRestAutoConfiguration"
+                "internal.org.springframework.content.rest.boot.autoconfigure.ContentRestAutoConfiguration",
+                "org.springframework.boot.autoconfigure.data.rest.RepositoryRestMvcAutoConfiguration"
```
vierbergenlars, line 38: Since this one is on the classpath, you could reference it using a class (that helps against accidental typos)

rschev: I can't have two `@AutoConfigureAfter`s on the same class.

vierbergenlars: No, but I wonder if you can set both `value` and `name` in an `@AutoconfigureAfter`:

```java
@AutoconfigureAfter(
   name={
       "internal.org.springframework..."
   },
   value={
      RepositoryRestMvcAutoConfiguration.class
   }
)
```

vierbergenlars (approved):

PR #120, opened 2023-09-01
Ensure that relations that are not exported are not shown in the create and edit forms

vierbergenlars:


tgeens (approved):

contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/PromotionCampaign.java
```diff
@@ -32,13 +32,12 @@ public class PromotionCampaign {
 
     @Column(updatable = false, nullable = false)
     @CollectionFilterParam(value = "promo_code")
-    @RestResource(rel = "d:promo-code")
```
vierbergenlars, line 35: Note: this was just a plain incorrect annotation that I cleaned up: promoCode is not a relation, so can't be a `@RestResource`

PR #122, opened 2023-09-05
Add HAL-FORMS templates for modifying relations

vierbergenlars:
- For to-one relations, the hal-forms templates on the entity allow both to set a relation and to clear a relation
- For to-many relations, we can only add a template to add items to a relation, as removing items from a relation must happen on a per-item basis

For now we leave adding a template to remove an item for a to-many relation out of scope.

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/links/SpringDataAssociationLinkCollector.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/links/SpringDataAssociationLinkCollector.java
@@ -70,14 +70,19 @@ class SpringDataAssociationLinkCollector implements ContentGridLinkCollector {
 
     private Link addAssociationAffordance(Link associationLink, Class<?> owner, Property association) {
         if (!association.getTypeInformation().isCollectionLike()) {
-            return Affordances.of(associationLink)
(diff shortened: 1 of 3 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/links/SpringDataAssociationLinkCollector.java
```diff
@@ -57,12 +68,84 @@ public Links getLinksForNested(Object object, Links existing) {
         return existing;
     }
 
-    private String readLinkName(PersistentProperty<?> property) {
-        var jsonProperty = property.findAnnotation(JsonProperty.class);
-        if(jsonProperty != null && !Objects.equals(jsonProperty.value(), JsonProperty.USE_DEFAULT_NAME)) {
-            return jsonProperty.value();
+    private Link addAssociationAffordance(Link associationLink, Class<?> owner, Property association) {
+        if (!association.getTypeInformation().isCollectionLike()) {
```
tgeens, line 72: there is a third case: `association.getTypeInformation().isMap()`

I know we don't generate anything with that datatype right now, but it might be prudent to add an else-if and throw ? (the survey-POC used a map)

tgeens (changes requested):

tgeens (approved):

PR #124, opened 2023-09-18
Fix database isolation in tests

vierbergenlars:
It turns out that the jdbc-based testcontainers has a problem where all
connections that use the same jdbc url are shared to the same database
instance. This causes a problem when tests run in parallel, as hibernate
sets up and destroys the schema on every application context start, but
it's all on the same database.

Using the spring `@ServiceConnection` instead ensures that a different
database is started for every time the application context is
(re)created, avoiding any problems with a shared database

### DIFF
--- a/contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/Invoice.java
+++ b/contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/Invoice.java
@@ -92,7 +92,7 @@ public class Invoice {
     private Customer counterparty;
 
     @OneToMany
-    @JoinColumn(name = "__invoice_id__orders", foreignKey = @ForeignKey(foreignKeyDefinition = "foreign key (\"invoice\") references \"invoice\" ON DELETE set NULL"))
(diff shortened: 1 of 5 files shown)


contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/Invoice.java
```diff
@@ -92,7 +92,7 @@ public class Invoice {
     private Customer counterparty;
 
     @OneToMany
-    @JoinColumn(name = "__invoice_id__orders", foreignKey = @ForeignKey(foreignKeyDefinition = "foreign key (\"invoice\") references \"invoice\" ON DELETE set NULL"))
+    @JoinColumn(name = "__invoice_id__orders", foreignKey = @ForeignKey(foreignKeyDefinition = "foreign key (\"__invoice_id__orders\") references \"invoice\" ON DELETE set NULL"))
```
rschev, line 95: Should this be part of this PR?

vierbergenlars: This is also a fix, before, the foreign key was not set up correctly (and hibernate ignored the incorrect definition, just logging a warning to the console all the time)

PR #125, opened 2023-09-18
Use jakarta bean validation on entities

vierbergenlars:
Use spring validator to run validation on entities before they are saved to the database

This allows giving clearer errors to users when something is wrong with the submitted data.

Previously, they would always receive a database error when trying to create an entity with a required attribute (or relation) missing.
The database error also only contains information about the first field that is incorrect.

Now, they receive a validation error that contains information about all fields that are required and missing.


contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/support/Content.java
```diff
@@ -28,6 +30,7 @@ public class Content {
 
     @MimeType
     @CollectionFilterParam(value = "mimetype")
+    @Pattern(regexp = "^(application|audio|font|image|model|text|video)/[a-z0-9!#$&\\-^_.+]+$", flags = Flag.CASE_INSENSITIVE)
```
rschev, line 33: According to StackOverflow you're missing [a few possible types](https://stackoverflow.com/questions/25201083/regex-to-match-and-validate-internet-media-type/48046041#48046041) but I'm not sure we need to do it this strictly anyway.

rschev (approved):

contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/support/Content.java
vierbergenlars: Oh, I know.
This is just a test regex pattern, not the final one that may actually be generated and used in applications.

PR #126, opened 2023-09-18
Use RFC7807 problem details for reporting errors

vierbergenlars:
Using [problem details](https://www.rfc-editor.org/rfc/rfc7807) for reporting errors to the client allows for returning a more specific error to the client than HTTP status codes allow.

Clients can use the specific error codes to show a proper message to end-users in the UI, or to autonomously take action to rectify the cause of the error.

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/mapping/PlainMapping.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/mapping/PlainMapping.java
@@ -17,5 +17,5 @@ import org.springframework.beans.factory.annotation.Qualifier;
 @Documented
 @Qualifier
 public @interface PlainMapping {
-
+    boolean followingAssociations() default false;
 }
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/mapping/PlainMapping.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/mapping/PlainMapping.java
@@ -9,13 +9,13 @@ import java.lang.annotation.Target;
 import org.springframework.beans.factory.annotation.Qualifier;
 
 /**
- * Qualifier annotation for {@link DomainTypeMapping} that relates to search query parameters
+ * Qualifier annotation for {@link DomainTypeMapping} that relates to a plain mapping that does not change property names
  */
 @Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE, ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
 @Documented
 @Qualifier
-public @interface SearchMapping {
+public @interface PlainMapping {
 
 }
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
@@ -38,14 +38,13 @@ public class ContentGridExceptionHandler {
 
     @ExceptionHandler(ConstraintViolationException.class)
     ResponseEntity<Problem> handleConstraintViolation(ConstraintViolationException exception) {
-        var type = switch (exception.getSQLState()) {
-            case "23505" -> ProblemType.UNIQUE_CONSTRAINT_VIOLATION;
-            default -> ProblemType.CONSTRAINT_VIOLATION;
+        var problem = switch (exception.getSQLState()) {
+            case "23505", "23P01" -> problemFactory.createProblem(ProblemType.DUPLICATE_VALUE, exception.getConstraintName())
+                    .withStatus(HttpStatus.CONFLICT);
+            default -> problemFactory.createProblem(ProblemType.CONSTRAINT_VIOLATION, exception.getConstraintName())
+                    .withStatus(HttpStatus.BAD_REQUEST);
         };
-        return responseEntityFactory.createResponse(
-                problemFactory.createProblem(type, exception.getConstraintName())
-                        .withStatus(HttpStatus.CONFLICT)
-        );
+        return responseEntityFactory.createResponse(problem);
     }
 
     @ExceptionHandler
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
@@ -1,6 +1,7 @@
 package com.contentgrid.spring.data.rest.problem;
 
 import com.contentgrid.spring.data.rest.problem.ext.ConstraintViolationProblemProperties;
+import com.contentgrid.spring.data.rest.problem.ext.ConstraintViolationProblemProperties.FieldViolationProblemProperties;
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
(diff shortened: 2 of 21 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/mapping/PlainMapping.java
```diff
@@ -9,13 +9,13 @@
...
+ * Qualifier annotation for {@link DomainTypeMapping} that relates to a plain mapping that does not change property names
  */
 @Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE, ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
 @Documented
 @Qualifier
-public @interface SearchMapping {
-
+public @interface PlainMapping {
```
tgeens, line 19: hmm, I was already looking for Spring MVC docs to figure out what the hell `@PlainMapping` was (like a `@GetMapping` variant I did not know about)

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
```diff
@@ -0,0 +1,189 @@
...
+
+    public static ValidationConstraintViolationMatcher validationConstraintViolation() {
+        return new ValidationConstraintViolationMatcher(List.of());
+    }
+
+    @With
+    @AllArgsConstructor
+    public static class ProblemDetailsMatcher implements ResultMatcher {
+
+        private final static String SENTINEL = "\0";
```
tgeens, line 47: what ? why ?

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
```diff
@@ -0,0 +1,189 @@
...
+            readProblemDetail(result);
+        }
+    }
+
+    @AllArgsConstructor
+    public static class ValidationConstraintViolationMatcher implements ResultMatcher {
+
+        private final static ProblemDetailsMatcher PROBLEM_DETAILS_MATCHER = new ProblemDetailsMatcher()
+                .withStatusCode(HttpStatus.BAD_REQUEST)
+                .withType("https://contentgrid.cloud/problems/integrity/validation-constraint-violation");
```
tgeens, line 102: Shouldn't we extract this url (or at least the base) into a dedicated class ?

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
vierbergenlars: This is a matcher used for tests.

It needs to distinguish between:

- Value not configured (don't perform a match) -> `SENTINEL`
- Value configured to null (match to null/absence of value) -> `null`
- Value configured to a string (match to that string)

We need some kind of value that can be used for the first case.
The string `\0` is very unlikely to be used in any problem details response, so I used that.

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
vierbergenlars: The URL should not change, because it is the identity of the problem, right?

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    @NonNull
+    private final MessageSourceAccessor messageSourceAccessor;
+
+    @NonNull
+    private final JsonPropertyPathConverter jsonPropertyPathConverter;
+
+    @ExceptionHandler(ConstraintViolationException.class)
+    @ResponseStatus(HttpStatus.CONFLICT)
+    Problem handleConstraintViolation(ConstraintViolationException exception) {
+        var type = switch (exception.getSQLState()) {
```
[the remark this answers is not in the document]
vierbergenlars: Unfortunately, this is the only way (short of string parsing an exception message) to determine that the problem is a unique constraint violation (which seems an important enough subtype to me to warrant a separate problem type)

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            case "23505" -> ProblemType.UNIQUE_CONSTRAINT_VIOLATION;
+            default -> ProblemType.CONSTRAINT_VIOLATION;
+        };
+        return problemFactory.createProblem(type, exception.getConstraintName())
+                .withStatus(HttpStatus.CONFLICT);
```
rschev, line 46: I don't think Conflict is appropriate for constraint violations that _aren't_ a unique constraint violation.

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
rschev: It might be nice to be able to refer to it with IDE autocomplete, rather than having to carefully copy-paste it all the time.

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
tgeens: So there was a semi-legitimate use for :hankey: and you did not go for that ? I'm disappointed now.

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
tgeens: :point_up:

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
vierbergenlars: I agree. However, I'm not sure what status to assign to which class exactly.

From [the Postgres docs](https://www.postgresql.org/docs/current/errcodes-appendix.html)


|Class 23| Integrity Constraint Violation | HTTP status code |
| ----------| ------------ | ---------- |
| 23000 | integrity_constraint_violation | :question: |
| 23001 | restrict_violation | :question: |
| 23502 | not_null_violation | 400 Bad Request |
| 23503 | foreign_key_violation | :question: |
| 23505 | unique_violation | 409 Conflict | 
| 23514 | check_violation | :question: |
| 23P01 | exclusion_violation | 409 Conflict |

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
vierbergenlars: Well, in the test class, the prefix is there in a constant, which is the one you would mostly use.

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
tgeens: In case of doubt: plain HTTP 400

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/problem/ProblemDetailsMockMvcMatchers.java
tgeens: :shrug:

rschev (approved):

PR #130, opened 2023-09-20
Update validation & unique constraint problem details type to reference that it is an input problem

vierbergenlars:


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ProblemType.java
```diff
@@ -6,9 +6,9 @@
 import org.springframework.context.MessageSourceResolvable;
 
 public enum ProblemType implements ProblemTypeResolvable {
-    VALIDATION_CONSTRAINT_VIOLATION("integrity", "validation-constraint-violation"),
+    INPUT_VALIDATION( "input", "validation"),
```
tgeens, line 9: why the extra space ?

```suggestion
    INPUT_VALIDATION("input", "validation"),
```

tgeens (approved):

PR #146, opened 2023-11-27
Make order of properties consistent in 'search' hal template

rschev:


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/querydsl/mapping/CollectionFiltersFactory.java
```diff
@@ -33,13 +34,9 @@ class CollectionFiltersFactory {
...
-        return new CollectionFiltersImpl(filters().collect(Collectors.toUnmodifiableMap(
-                CollectionFilter::getFilterName,
-                Function.identity(),
-                (a, b) -> {
-                    throw new IllegalStateException("Duplicate filter name '%s' mapping to paths '%s' and '%s'".formatted(a.getFilterName(), a.getPath(), b.getPath()));
-                }
-        )));
+        LinkedHashMap<String, CollectionFilter<?>> orderedMap = new LinkedHashMap<>();
+        filters().forEachOrdered(filter -> orderedMap.put(filter.getFilterName(), filter));
+        return new CollectionFiltersImpl(orderedMap);
```
vierbergenlars, line 39: 1. This is no longer an unmodifiable map
2. There is a `Collector` where you can supply the map constructor: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Collectors.html#toMap(java.util.function.Function,java.util.function.Function,java.util.function.BinaryOperator,java.util.function.Supplier)
3. This doesn't check for duplicate filter names

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/querydsl/mapping/CollectionFiltersImpl.java
```diff
@@ -1,15 +1,14 @@
...
-import java.util.Map;
+import java.util.LinkedHashMap;
 import java.util.Optional;
 import java.util.stream.Stream;
 import lombok.RequiredArgsConstructor;
 
 @RequiredArgsConstructor
 class CollectionFiltersImpl extends AbstractCollectionFiltersImpl {
-    private final Map<String, CollectionFilter<?>> filters;
+    private final LinkedHashMap<String, CollectionFilter<?>> filters;
```
vierbergenlars, line 11: Does this type have to be changed here?

rschev: My reasoning was that if you want it to actually be ordered, then it should be changed in the class. But I guess that since this class is package-private, we're relying on nothing but the factory initializing it?

tgeens (approved):

PR #151, opened 2023-12-01
Add multi-tenant oauth2 resource server auto-configuration

tgeens:


contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/security/MultiTenantOAuth2ResourceServerAutoConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+@AutoConfiguration(before = { OAuth2ResourceServerAutoConfiguration.class, SecurityAutoConfiguration.class })
+@EnableConfigurationProperties(MultiTenantOAuth2Properties.class)
+@ConditionalOnClass(value = {JwtIssuerAuthenticationManagerResolver.class, SecurityFilterChain.class, HttpSecurity.class})
+@ConditionalOnWebApplication(type = Type.SERVLET)
+class MultiTenantOAuth2ResourceServerAutoConfiguration {
+
+
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnProperty("contentgrid.security.oauth2.trusted-jwt-issuers[0]")
```
vierbergenlars, line 28:
```suggestion
    @ConditionalOnProperty("contentgrid.security.oauth2.trusted-jwt-issuers")
```

Doesn't this work the same way?

Or do we only want to enable this when there is at least one trusted issuer?

vierbergenlars: Also, shouldn't this whole class be conditional on that property, given that `jwtIssuerSecurityFilterChain` is dependent on this bean being created?

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/security/MultiTenantOAuth2ResourceServerAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    JwtIssuerAuthenticationManagerResolver authenticationManagerResolver(MultiTenantOAuth2Properties oauth2Properties) {
+        return new JwtIssuerAuthenticationManagerResolver(oauth2Properties.getTrustedJwtIssuers());
+    }
+
+    @Bean
+    @ConditionalOnBean(JwtIssuerAuthenticationManagerResolver.class)
+    SecurityFilterChain jwtIssuerSecurityFilterChain(HttpSecurity http,
+            AuthenticationManagerResolver<HttpServletRequest> authManagerResolver) throws Exception {
```
vierbergenlars, line 36: Conditional on a bean of a certain type being present, but the object that's injected is a generic interface.

That means that it is possible for a different `AuthenticationManagerResolver` to be injected here, which would be unexpected. (multiple beans of the same type won't result in an error if one of them is marked `@Primary`)

tgeens: fair, I'll specialise the argument so they match

vierbergenlars (approved):

PR #156, opened 2023-12-12
Setup contentgrid-spring-swagger

tgeens:
New small contentgrid-spring subproject with 2 goals:
* facilitate updating `swagger-ui` dependency with renovate
* get rid of the following dependencies in the ContentGrid apps `build.gradle`:
  - `org.webjars:swagger-ui`
  - `org.webjars:webjars-locator-core`


settings.gradle
```diff
@@ -25,4 +25,5 @@ include 'contentgrid-spring-boot-starter-annotations'
 include 'contentgrid-spring-boot-platform'
 include 'contentgrid-spring-data-rest'
 include 'contentgrid-spring-integration-events'
-include 'contentgrid-spring-querydsl'
\ No newline at end of file
+include 'contentgrid-spring-querydsl'
+include 'contentgrid-spring-swagger'
```
vierbergenlars, line 29:
```suggestion
include 'contentgrid-spring-swagger-ui'
```

This only supports swagger UI, not the open api spec (a long time ago called swagger spec) itself.

contentgrid-spring-swagger/src/main/resources/swagger-initializer.js
```diff
@@ -0,0 +1,20 @@
+window.onload = function() {
+    //<editor-fold desc="Changeable Configuration Block">
+
+    // the following lines will be replaced by docker/configurator, when it runs in a docker-container
```
vierbergenlars, line 4: This is not exactly true; that's a comment coming from where the file was copied from, bus is not accurate in this project

contentgrid-spring-swagger/src/main/java/com/contentgrid/spring/swaggerui/SwaggerUIInitializerController.java
```diff
@@ -0,0 +1,31 @@
+package com.contentgrid.spring.swaggerui;
```
vierbergenlars, line 1: Isn't the resource file served automatically when it's put on the classpath in the correct location?

tgeens: It is, but in this case it is problematic:
- either you put it in a versioned folder, like /webjars/swagger-ui/4.11.1/ and you have problems when the dependency gets updated
- if you put it one level higher, the webjar version detection breaks

vierbergenlars: That is something useful to put in a comment on the class then?

contentgrid-spring-swagger/src/main/resources/swagger-initializer.js
tgeens: I copied it varbatim from the current project. I'll clean it up.

contentgrid-spring-swagger/src/main/java/com/contentgrid/spring/swaggerui/SwaggerUIInitializerController.java
tgeens: See updated javadoc

vierbergenlars (approved):

PR #158, opened 2023-12-12
Fix more sonar smells

tgeens:


contentgrid-spring-integration-events/src/main/java/com/contentgrid/spring/integration/events/EntityChangeEventTransformer.java
```diff
@@ -19,7 +20,7 @@ public class EntityChangeEventTransformer extends AbstractPayloadTransformer<Ent
     @Override
     protected ChangeEventPayload transformPayload(EntityChangeEvent changeEvent) {
         return new ChangeEventPayload(
-                changeEvent.getTrigger(),
+                changeEvent.getTrigger().toString().toLowerCase(Locale.ROOT),
```
vierbergenlars, line 23: This should either also use `.name()`, like in `EntityChangeEventPublisher`, or both should be using `toString()`

I prefer using `.name()`, as it is more obvious to me that it's the enum name. For `toString()`, I always have to remember that enum does not use the default `Object#toString()` implementation.

contentgrid-spring-querydsl/src/main/java/com/contentgrid/spring/querydsl/mapping/UnsupportedCollectionFilterPredicateException.java
```diff
@@ -6,12 +6,10 @@
 
 @Getter
 public class UnsupportedCollectionFilterPredicateException extends UnsupportedOperationException{
-    private final QuerydslPredicateFactory<?, ?> predicateFactory;
```
vierbergenlars, line 9: I'm not sure that this is unnecessary. It can be used to provide more details why this predicate does not work?

tgeens: Ok, this is not used - at this moment.

Then again, exceptions should be serializable. Would you prefer making this a transient field ?

tgeens: As discussed: we're removing the field.

vierbergenlars (approved):

PR #172, opened 2024-01-22
ACC-1159: Add publication of audit events to AMQP with CloudEvents

vierbergenlars:
- [ACC-1159] Add cloud events support for RabbitTemplate MessageConverter
- [ACC-1159] Add AMQP cloudevents publishing of audit events
- [ACC-1159] Add test for AMQP cloudEvents publishing

[ACC-1159]: https://xenitsupport.jira.com/browse/ACC-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tgeens (changes requested):

vierbergenlars:
Updated this PR to:

- First emit the audit event into the ApplicationContext
- Then, subscribe to the audit events emitted into the ApplicationContext and send it over a MessagingTemplate to the outside (transforming it to a CloudEvent, then to a binary representation with the CloudEventConverter)

contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/handler/messaging/AuditEventToCloudEventMessageConverter.java
```diff
@@ -0,0 +1,94 @@
...
+                    .build(),
+            AuditEventToCloudEventMapper.build(EntitySearchAuditEvent.class)
+                    .cloudEventType("cloud.contentgrid.audit.entity.list")
+                    .cloudEventSubject(AbstractAuditEvent::getRequestUri)
+                    .build()
+    );
+
+    @Override
+    public Object fromMessage(Message<?> message, Class<?> targetClass) {
+        // TODO: implement converting back from a message to audit event?
```
tgeens, line 63: rather throw an exception, isntead of silently returning `null` ?

contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/handler/messaging/AuditEventToCloudEventMessageConverter.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @NonNull
+    private final PojoCloudEventData.ToBytes<AbstractAuditEvent> mapper;
+
+    @NonNull
+    private final URI source;
+
+    private static final Set<AuditEventToCloudEventMapper> TYPE_MAPPING = Set.of(
+            AuditEventToCloudEventMapper.build(BasicAuditEvent.class)
+                    .cloudEventType("cloud.contentgrid.audit.basic")
```
tgeens, line 36: It's `cloud.contentgrid` prefiexed because it is runtime ?

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/ContentGrid.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.springframework.beans.factory.annotation.Qualifier;
+
+
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Qualifier
+public @interface ContentGrid {
```
tgeens, line 15: wait, what does this do ? :upside_down_face:

contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/handler/messaging/AuditEventToCloudEventMessageConverter.java
vierbergenlars: Yes, and the cloud event type should be in reverse-DNS format according to the cloud-events spec

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/ContentGrid.java
vierbergenlars: It's a bean qualifier annotation.

For the `MessageConverter`s, I am creating beans that implement a Spring framework interface. I don't want other places that also register beans to interfere with these.

contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/handler/messaging/AuditEventToCloudEventMessageConverter.java
vierbergenlars: When returning null, another message converter gets a chance to convert it. If no message converters return an object, an exception is thrown from somewhere.

tgeens: Ok - what should happen with this `// TODO` ? When would we need the conversion in the other direction ?

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/ContentGrid.java
tgeens: Discussed: renaming to `@ContentGridMessaging` and moving to the `.messaging` package would be more appropriate

tgeens (approved):

PR #178, opened 2024-02-05
Add system properties to audit event [ACC-1196]

rschev:
System properties are applicationId and deploymentId.

### DIFF
--- a/contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/extractor/BasicAuditEventExtractor.java
+++ b/contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/extractor/BasicAuditEventExtractor.java
@@ -6,12 +6,16 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Optional;
 import java.util.Set;
+import lombok.RequiredArgsConstructor;
 import org.springframework.core.Ordered;
 import org.springframework.http.server.observation.ServerRequestObservationContext;
 import org.springframework.web.servlet.HandlerMapping;
 import org.springframework.web.servlet.resource.ResourceHttpRequestHandler;
 
+@RequiredArgsConstructor
 public class BasicAuditEventExtractor implements AuditEventExtractor {
+    private final String appId;
+    private final String deplId;
 
     private static final Set<HandlerMethodMatcher> IGNORED_HANDLERS = Set.of(
             HandlerMethodMatcher.builder()
@@ -62,7 +66,9 @@ public class BasicAuditEventExtractor implements AuditEventExtractor {
                 .requestMethod(context.getCarrier().getMethod())
                 .requestUri(context.getCarrier().getRequestURI())
                 .responseStatus(context.getResponse().getStatus())
-                .responseLocation(location.orElse(null));
+                .responseLocation(location.orElse(null))
+                .applicationId(appId)
+                .deploymentId(deplId);
     }
 
     @Override
--- a/contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/extractor/BasicAuditEventExtractor.java
+++ b/contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/extractor/BasicAuditEventExtractor.java
@@ -12,10 +12,7 @@ import org.springframework.http.server.observation.ServerRequestObservationConte
 import org.springframework.web.servlet.HandlerMapping;
 import org.springframework.web.servlet.resource.ResourceHttpRequestHandler;
 
-@RequiredArgsConstructor
 public class BasicAuditEventExtractor implements AuditEventExtractor {
-    private final String appId;
-    private final String deplId;
 
     private static final Set<HandlerMethodMatcher> IGNORED_HANDLERS = Set.of(
             HandlerMethodMatcher.builder()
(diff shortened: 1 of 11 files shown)


contentgrid-spring-audit-logging/src/main/java/com/contentgrid/spring/audit/extractor/BasicAuditEventExtractor.java
```diff
@@ -6,6 +6,7 @@
 import java.net.URISyntaxException;
 import java.util.Optional;
 import java.util.Set;
+import lombok.RequiredArgsConstructor;
```
vierbergenlars, line 9: Unused import

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/audit/ContentGridAuditMessagingAutoConfiguration.java
```diff
@@ -4,8 +4,8 @@
 import com.contentgrid.spring.audit.handler.messaging.AuditEventToCloudEventMessageConverter;
 import com.contentgrid.spring.audit.handler.messaging.Jackson2AuditMessagingModule;
 import com.contentgrid.spring.audit.handler.messaging.MessageSendingAuditHandler;
-import com.contentgrid.spring.boot.autoconfigure.messaging.ContentGridMessaging;
 import com.contentgrid.spring.boot.autoconfigure.audit.ContentGridAuditMessagingAutoConfiguration.ContentGridAuditMessagingProperties;
+import com.contentgrid.spring.boot.autoconfigure.messaging.ContentGridMessaging;
```
vierbergenlars, line 8: Unused import

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/messaging/ContentGridMessagingSystemProperties.java
```diff
@@ -0,0 +1,10 @@
+package com.contentgrid.spring.boot.autoconfigure.messaging;
+
+import java.util.UUID;
+import lombok.Data;
+
+@Data
+public class ContentGridMessagingSystemProperties {
+    private String deploymentId = new UUID(0, 0).toString();
+    private String applicationId = new UUID(0, 0).toString();
```
vierbergenlars, line 9: Shouldn't we leave them set to null if they are not set?
And then don't send the application/deployment id headers?

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/messaging/ContentGridPropertiesMessageConverter.java
```diff
@@ -0,0 +1,45 @@
...
+    public Message<?> toMessage(Object payload, MessageHeaders headers) {
+        return delegate.toMessage(payload, extendHeaders(headers));
+    }
+
+    private MessageHeaders extendHeaders(MessageHeaders headers) {
+        return MessageBuilder.withPayload("foo").copyHeaders(headers)
+                .setHeaderIfAbsent("applicationId", systemProperties.getApplicationId())
+                .setHeaderIfAbsent("deploymentId", systemProperties.getDeploymentId())
+                .build()
+                .getHeaders();
```
vierbergenlars, line 43: `MessageHeaders` is mutated everywhere; so why the effort to make them immutable?

rschev: From the javadoc of MessageHeaders:
> **IMPORTANT**: This class is immutable. Any mutating operation such as `put(..)`, `putAll(..)` and others will throw `UnsupportedOperationException`.
> Subclasses do have access to the raw headers, however, via `getRawHeaders()`.
> One way to create message headers is to use the `MessageBuilder`:
> `MessageBuilder.withPayload("foo").setHeader("key1", "value1").setHeader("key2", "value2");`
> ...

What way to mutate it would you like me to use?

vierbergenlars: There are 2 options

- `MessageHeaders` implements `Map`, so you could copy that to a mutable `Map`, insert the additional headers and then create a new `MessageHeaders` from that map.
- Use the `MessageHeaderAccessor` to mutate the headers and return a new `MessageHeaders`.

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/messaging/ContentGridPropertiesMessageConverter.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public Message<?> toMessage(Object payload, MessageHeaders headers) {
+        return delegate.toMessage(payload, extendHeaders(headers));
+    }
+
+    private MessageHeaders extendHeaders(MessageHeaders headers) {
+        if (systemProperties.getApplicationId() == null && systemProperties.getDeploymentId() == null) {
+            return headers;
+        }
```
vierbergenlars, line 43: This isn't really necessary, I think?
Just trying to avoid some copying?

vierbergenlars (approved):

PR #190, opened 2024-02-29
Add auditing metadata in tests

NielsCW:


contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/test/fixture/invoicing/InvoicingApplicationTests.java
```diff
@@ -138,15 +150,28 @@ void doInTransaction(ThrowingCallable callable) {
...
+                .withZone(ZoneId.of("GMT"));
+        return formatter.format(date);
+    }
+
+    public static void loginUser(String username) {
+        var user = new User(username, "password", Set.of());
+        var authentication = new UsernamePasswordAuthenticationToken(user, user.getPassword(), user.getAuthorities());
+        var context = SecurityContextHolder.getContext();
+        context.setAuthentication(authentication);
+    }
```
tgeens, line 164: can you explain how this is related to `@WithMockUser` on the class-level ?

NielsCW: `@WithMockUser` sets the `SecurityContext` to contain an authenticated `User` with the given username. This `loginUser()` overrides `SecurityContext.authentication` to contain a new authenticated `User`. So that following requests are send by the new user. This is used to test whether `lastModifiedBy` can be altered when a different user makes a PUT request for example.

tgeens: so the "problem" is that if you override `@WithMockUser("other")` on the test-method level, the before-each also runs with user `"other"` ?

NielsCW: With `@WithMockUser` alone, you can only test one user per test:
If a test is marked with `@WithMockUser("John")` and it calls a function that modifies an object in a repository with `@WithMockUser("Bob")`, `lastModifiedDate` of that object will be updated, but `lastModifiedBy` will contain `"John"`.

tgeens: and something like this does not work: https://docs.spring.io/spring-security/reference/servlet/test/mockmvc/authentication.html#test-mockmvc-securitycontextholder-rpp ?

tgeens:
We need to figure out what the implications are of putting `spring-security` and `spring-boot-starter-oauth2-resource-server` on the classpath (in a separate PR)

contentgrid-spring-audit-logging/src/test/java/com/contentgrid/spring/audit/AuditObservationHandlerTest.java
```diff
@@ -434,9 +437,13 @@ void contentDelete() throws Exception {
...
     class HandlesThrownExceptions {
 
-        private static final ServerRequestObservationContext SERVER_REQUEST_OBSERVATION_CONTEXT = new ServerRequestObservationContext(
-                new MockHttpServletRequest(),
-                new MockHttpServletResponse());
+        private static final ServerRequestObservationContext SERVER_REQUEST_OBSERVATION_CONTEXT = createContext();
+
+        private static ServerRequestObservationContext createContext() {
+            var request = new MockHttpServletRequest();
+            request.setAttribute(HandlerMapping.BEST_MATCHING_HANDLER_ATTRIBUTE, "true");
```
tgeens, line 444: why is this necessary ?

NielsCW: The tests used to work because the `handler` in `BasicAuditEventExtractor.createEventBuilder` is `null` and it was missing a null check. Since the null-check is added and the requests are crafted manually, we add this attribute to make the handler non-null and the request isn't ignored.

tgeens: isn't it fishy that all the code expect that it contains a request-handler-method and we're putting the string `"true"` in there ? :thinking:

tgeens (approved):

PR #191, opened 2024-03-05
Make cg:entity, cg:relation & cg:content links always an array

vierbergenlars:
HAL specifies that links should not switch between a list and a single value.
`cg:entity`, `cg:relation` & `cg:content` are always used in a context where there are potentially multiple entries present.

Because this feature uses the new `MediaTypeConfigurationCustomizer` from spring-hateoas 2.2, **support for Spring Boot 3.1 is dropped.**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - spring-boot-version: 3.1.+
+          - spring-boot-version: 3.2.+
             experimental: false
           - spring-boot-version: 3.+
             experimental: true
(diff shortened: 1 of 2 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/links/ContentGridSpringDataLinksConfiguration.java
```diff
@@ -45,6 +49,14 @@ CurieProviderCustomizer contentGridCurieProviderCustomizer() {
         return CurieProviderCustomizer.register(ContentGridLinkRelations.CURIE, ContentGridLinkRelations.TEMPLATE);
     }
 
+    @Bean
+    MediaTypeConfigurationCustomizer<HalConfiguration> contentGridLinksMediaTypeConfigurationCustomizer() {
+        return halConfiguration -> halConfiguration
+                .withRenderSingleLinksFor(ContentGridLinkRelations.CONTENT, RenderSingleLinks.AS_ARRAY)
+                .withRenderSingleLinksFor(ContentGridLinkRelations.RELATION, RenderSingleLinks.AS_ARRAY)
+                .withRenderSingleLinksFor(ContentGridLinkRelations.ENTITY, RenderSingleLinks.AS_ARRAY);
```
tgeens, line 57: @michaelXenit heads up - those links are always going to be arrays in the future, even if there is a single _content_-attribute

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - spring-boot-version: 3.1.+
```
tgeens, line 28: very sneaky

tgeens (approved):

PR #192, opened 2024-03-05
Use new MediaTypeConfigurationCustomizer to customize HalFormsConfiguration

vierbergenlars:
The MediaTypeConfigurationCustomizer is a new way to configure HalFormsConfiguration without creating the whole configuration in one place.
Since we only need to configure hal-forms options on some fields, it is a good candidate to be moved to the MediaTypeConfigurationCustomizer approach.
 
Also add tests that check that the options are added correctly to relations.


tgeens (commented):
Good in principle, but this needs to be blocked until cg-apps are upgraded to spring boot 3.2 ?

PR #200, opened 2024-03-22
Add spring-boot-starter-oauth2-resource-server to contentgrid-spring-data-rest tests

NielsCW:


contentgrid-spring-data-rest/build.gradle
```diff
@@ -42,7 +42,10 @@ dependencies {
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
     testImplementation 'org.springframework.boot:spring-boot-starter-web'
     testImplementation 'jakarta.transaction:jakarta.transaction-api'
+    testImplementation 'org.springframework.security:spring-security-test'
+    testImplementation 'org.springframework.boot:spring-boot-starter-oauth2-resource-server'
     testImplementation project(':contentgrid-spring-boot-autoconfigure')
+    testImplementation project(':contentgrid-spring-data-support')
```
vierbergenlars, line 48: It doesn't look like you're using any classes from this project in your tests. Do you need this dependency?

NielsCW: Not now in this PR, it is needed when adding audit metadata tests

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/test/security/WithMockJwt.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.security.test.context.support.WithSecurityContext;
+import org.springframework.security.test.context.support.WithSecurityContextFactory;
+import org.springframework.util.StringUtils;
+
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Documented
+@WithSecurityContext(factory = WithMockJwt.WithMockJwtSecurityContextFactory.class)
+@Import({SkipCsrfConfiguration.class})
```
tgeens, line 29: Now I'm wondering what happens with this import on the method-level ? :thinking:
That won't work, right ?

NielsCW: Removing the `@WithMockJwt` on the class and placing it on the method testing POST -> test fails with http 403

vierbergenlars: If it's not supported on methods, then `@Target` should not include `ElementType.METHOD`

tgeens: but it is intended to be used on method level too ...

NielsCW: Setting the `@WithMockJwt` on the method overrides the properties of the one on the class. It is when the `@WithMockJwt` is missing on the class that the `SkipCsrfConfiguration` does not get loaded.

tgeens: In that case, you might need to use `@Import(SkipCsrfConfiguration.class)` on the test-class first. 

Ok, fine with me.

tgeens, line 29:
```suggestion
@Import(SkipCsrfConfiguration.class)
```

contentgrid-spring-data-support/build.gradle
```diff
@@ -24,7 +24,7 @@ dependencies {
     implementation 'jakarta.persistence:jakarta.persistence-api'
     implementation 'com.fasterxml.jackson.core:jackson-annotations'
     implementation 'org.springframework.data:spring-data-jpa'
-    compileOnly 'org.springframework.security:spring-security-oauth2-jose'
+    implementation 'org.springframework.security:spring-security-oauth2-jose'
```
tgeens, line 27: why is this change required ?

tgeens (approved):

PR #206, opened 2024-03-28
Audit-metadata: fix null date fields when spring security is missing from classpath

NielsCW:
Before:
image
After:
image


contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/data/audit/JpaAuditingAutoConfiguration.java
```diff
@@ -1,18 +1,27 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.data.domain.AuditorAware;
+import org.springframework.data.jpa.repository.config.EnableJpaAuditing;
 import org.springframework.security.oauth2.jwt.JwtClaimAccessor;
 
 @AutoConfiguration(after = {HibernateJpaAutoConfiguration.class})
 @ConditionalOnBean(EntityManagerFactory.class)
-@ConditionalOnClass({JpaAuditingConfiguration.class, JwtClaimAccessor.class})
-@Import(JpaAuditingConfiguration.class)
+@ConditionalOnClass(AuditMetadata.class)
```
vierbergenlars, line 18: This is not really dependent on the presence of the `AuditMetadata` class, only the `jwtAuditorAware` is dependent on the `UserMetadata` class being present.

Since `UserMetadata` is part of this class signature, you can't put `@ConditionalOnClass()` on the method, as the class would still attempt to get loaded.

Instead, I think 2 separate classes are necessary:
 - one to enable JPA Auditing; unconditionally. That can just be this class
 - one to create the `jwtAuditorAware` bean; conditional on `UserMetadata` and `JwtClaimAccessor` being on the classpath

These don't have to be autoconfigurations or separate files, the class for `jwtAuditorAware` can be a nested `@Configuration` class that is `@Import`ed by the autoconfiguration.

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/data/audit/JpaAuditingAutoConfiguration.java
```diff
@@ -1,18 +1,33 @@
...
 @AutoConfiguration(after = {HibernateJpaAutoConfiguration.class})
 @ConditionalOnBean(EntityManagerFactory.class)
-@ConditionalOnClass({JpaAuditingConfiguration.class, JwtClaimAccessor.class})
-@Import(JpaAuditingConfiguration.class)
+@EnableJpaAuditing
+@Import(JwtAuditorAwareConfiguration.class)
 public class JpaAuditingAutoConfiguration {
 
+    @Configuration
+    @ConditionalOnClass({UserMetadata.class, JwtClaimAccessor.class})
```
vierbergenlars, line 25: So, there will be no auditor when the `JwtClaimsAccessor` is not on the classpath.

So, when `spring-security-oauth2-jose` (dependency of `spring-boot-starter-oauth2-resource-server`) is not on the classpath, there will be no auditor; also not for anonymous users?

NielsCW: There could be a default auditorAware in case when there is no `JwtClaimAccessor` on the classpath and there is still a `SecurityContextHolder` on the classpath.

contentgrid-spring-boot-autoconfigure/src/test/java/com/contentgrid/spring/boot/autoconfigure/data/audit/JpaAuditingAutoConfigurationTest.java
```diff
@@ -14,35 +15,48 @@
...
                     assertThat(context).doesNotHaveBean(JwtAuditorAware.class);
-                    assertThat(context).doesNotHaveBean(AuditingEntityListener.class);
+                    assertThat(context).doesNotHaveBean(DefaultAuditorAware.class);
+                    assertThat(context).hasSingleBean(AuditingEntityListener.class);
+                });
+    }
+
+    @Test
+    void checkWithoutJwtClaimAccessor() {
+        contextRunner.withInitializer(ConditionEvaluationReportLoggingListener.forLogLevel(LogLevel.INFO))
```
tgeens, line 53: if you are using the `ConditionEvaluationReportLoggingListener` everywhere anyway, you should probably assign it on the shared `contextRunner` instance ?

tgeens (approved):

PR #208, opened 2024-04-03
Move etag header tests to separate test file

NielsCW:
Move optimistic locking tests about `If-Match` and `ETag` headers from `InvoicingApplicationTest` to a new file

### DIFF
--- a/contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/etag/OptimisticLockingTest.java
+++ b/contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/etag/OptimisticLockingTest.java
@@ -27,6 +27,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang.StringUtils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -42,9 +43,7 @@ import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.web.servlet.MockMvc;
 
 @Slf4j
-@SpringBootTest(properties = {
-        "server.servlet.encoding.enabled=false" // disables mock-mvc enforcing charset in request
-})
+@SpringBootTest
 @ContextConfiguration(classes = {
         InvoicingApplication.class,
 })
@@ -261,6 +260,7 @@ public class OptimisticLockingTest {
         }
 
         @Test
+        @Disabled("ACC-1313")
         void postMultipartInvoiceAndContent_shouldSetETag_http201() throws Exception {
             var file = new MockMultipartFile("content", "content.txt", MIMETYPE_PLAINTEXT_UTF8,
                     UNICODE_TEXT.getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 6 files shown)


contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/etag/OptimisticLockingTest.java
```diff
@@ -0,0 +1,431 @@
...
+    void checkETagExists(String url) throws Exception {
+        mockMvc.perform(get(url))
+                .andExpect(status().isOk())
+                .andExpect(headers().etag().exists());
+    }
+
+    void checkETagUnchanged(String url, int original) throws Exception {
+        mockMvc.perform(get(url))
+                .andExpect(status().isOk())
+                .andExpect(headers().etag().isEqualTo(original));
```
tgeens, line 130: etag is a string, not an `int` ? this will never be equal ?

NielsCW: The `ETagHeaderMatcher` is a custom header matcher that expects an int value for `isEqualTo` and `isNotEqualTo`, should we expect `String` instead?

tgeens: Yes, `etag` is an ASCII string, not an `int`
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#etag_value

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/etag/OptimisticLockingTest.java
```diff
@@ -0,0 +1,433 @@
...
+        INVOICE_1_VERSION = invoice.getVersion();
+    }
+
+    void checkETagExists(String url) throws Exception {
+        mockMvc.perform(get(url))
+                .andExpect(status().isOk())
+                .andExpect(headers().etag().exists());
+    }
+
+    void checkETagUnchanged(String url, int original) throws Exception {
```
tgeens, line 127: I would also change the type in these arguments:
```suggestion
    void checkETagUnchanged(String url, String etag) throws Exception {
```

contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/test/matchers/ETagHeaderMatcher.java
```diff
@@ -9,22 +9,22 @@
 import org.springframework.test.web.servlet.ResultMatcher;
 
 @RequiredArgsConstructor(access = AccessLevel.PACKAGE)
-public class EtagHeaderMatcher {
+public class ETagHeaderMatcher {
 
-    public static String toEtag(int version) {
+    public static String toETag(int version) {
```
tgeens, line 14: `String.valueOf(version)`

tgeens: and if it's a `String` return type, better to just inline that function everywhere

NielsCW: Gives me the following exception in tests: `Request processing failed: java.lang.IllegalArgumentException: Could not parse header 'If-Match' with value '0'`, it needs the extra quotes to succeed.

NielsCW: Application responds to GET requests with `ETag:""0""`, so in the request matcher also needs the quotes since it calling `equals`

tgeens (approved):

PR #212, opened 2024-04-16
Use S3-storage instead of file system storage for storing content in tests

NielsCW:


tgeens (approved):

tgeens:
I don't understand why the integration tests for 3.2.4 succeed ?

PR #216, opened 2024-04-17
Fix validation in the case a newer release dropped the target entity of a required relation

NielsCW:
Add support to fix:
- GET requests to existing entities (with required relation removed) result in http 200 instead of http 404
- PUT requests to existing entities (with required relation removed) result in http 200 instead of http 400
- PATCH requests to existing entities (with required relation removed) result in http 200 instead of http 400

Not supported in this PR:
- POST request to create entity (with required relation removed) always fail with http 400

Side effects:
- PATCH request to existing entities where required relation was not removed now result in http 400 database constraint violation instead of http 400 validation failed when using an invalid relation payload (because PATCH doesn't ignore relations unlike PUT)

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/BeanValidationRepositoryEventListener.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/BeanValidationRepositoryEventListener.java
@@ -1,5 +1,6 @@
 package com.contentgrid.spring.data.rest.validation;
 
+import jakarta.validation.groups.Default;
 import lombok.RequiredArgsConstructor;
 import org.hibernate.Hibernate;
 import org.springframework.beans.factory.ObjectProvider;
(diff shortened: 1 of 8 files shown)


contentgrid-spring-data-rest/src/testFixtures/java/com/contentgrid/spring/test/fixture/invoicing/model/Order.java
```diff
@@ -34,7 +35,7 @@ public class Order {
     @JsonProperty(access = Access.READ_ONLY)
     private UUID id;
 
-    @ManyToOne
+    @ManyToOne(fetch = FetchType.LAZY)
```
tgeens, line 38: I don't think cg-apps have `FetchType.LAZY` for `@ManyToOne`  relations right now ? why are we changing this here ?

vierbergenlars:
This is _an_ approach to try to solve some compatibility problems when running an older application on a newer codebase, but it doesn't cover all cases.

I think we should investigate a different approach, by actually having the database compatibility shims behave more closely to the old behavior of the database (being: actually keeping the old data around for the old application)

PR #232, opened 2024-05-27
No content properties in application/json create form [ACC-1403]

rschev:


contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/webmvc/HalFormsProfileControllerWithoutMultipartTest.java
vierbergenlars, line 1: This test does not cover the same as `HalFormsProfileControllerTest`, it only covers the forms that are *currently* affected (or that we think are the only ones that are currently affected) by the property.

Other forms that should remain the same should be tested as well.
As we don't want to duplicate the implementation that does not change, you can create an abstract parent class that contains the tests for forms that do not change

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/webmvc/DefaultDomainTypeToHalFormsPayloadMetadataConverter.java
```diff
@@ -80,7 +80,7 @@ public PayloadMetadata convertToUpdatePayloadMetadata(Class<?> domainType) {
         )
                 .filter(property -> !Objects.equals(property.getInputType(), HtmlInputType.URL_VALUE))
                 .forEachOrdered(properties::add);
-        return new ClassnameI18nedPayloadMetadata(domainType, properties);
+        return new PayloadMetadataAndMediaType(new ClassnameI18nedPayloadMetadata(domainType, properties), getCreatePayloadMediaType());
```
vierbergenlars, line 83: This is not the correct mediatype for the update form?

The update form is always json, multipart upload is not supported for updates?

PR #236, opened 2024-06-11
[ACC-136] Update auditing metadata attributes and tests

NielsCW:
- Update database column names of audit metadata attributes to reflect what is generated
- Update setter of audit metadata attributes to reflect what is generated


vierbergenlars:
Can you split off the removal of the tests? It's not directly related to the audit metadata feature.

Additionally, I would like some tests that verify the correct behavior of the content endpoints. Since it's not related to audit metadata, maybe in a different class.

vierbergenlars (approved):

PR #243, opened 2024-06-26
[ACC-1460] add last-modified tests for content properties

NielsCW:
The rest controller of spring-content uses the file metadata instead of the entity `@LastModifiedBy` metadata for setting the `Last-Modified` header or checking the `If-Unmodified-Since` header. Therefore tests for checking these headers for content endpoints don't belong under `AuditMetadataTest` and are moved to `LastModifiedContentTest`


contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/auditing/ContentLastModifiedTest.java
vierbergenlars, line 1: Where are the tests for GET with `If-Modified-Since`? (if it has changed; the answer should be 200 OK, otherwise 304 Not Modified)

rschev (approved):

PR #244, opened 2024-06-28
[ACC-1453] use abac-context to filter automations shown to the user

NielsCW:
- move `AutomationsRestController` and related classes to a separate project: `contentgrid-automations-rest`
- The `AutomationsRestController` retrieves the ABAC Context to perform filtering on the `AutomationModel`s
- The `AutomationModelVisitor` implements the [`ThunkExpressionVisitor`](https://github.com/xenit-eu/thunx/blob/v0.11.0/thunx-model/src/main/java/com/contentgrid/thunx/predicates/model/ThunkExpressionVisitor.java) interface to resolve the ABAC Context for a given `AutomationModel`


contentgrid-automations-rest/src/main/java/com/contentgrid/automations/rest/AutomationModelVisitor.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+            case AND -> {
+                yield Scalar.of(terms.stream()
+                        .map(scalar -> scalar.assertResultType(Boolean.class))
+                        .map(ThunkExpression::maybeValue)
+                        .flatMap(Optional::stream)
+                        .map(Boolean.class::cast)
+                        .reduce(true, (left, right) -> left && right));
+            }
+            case OR -> {
```
rschev, line 49: Should we assert that there's _at least_ one term for this? Otherwise the reduce will just fall back to false and you won't actually see an exception telling you there's something wrong with the policy. Same for `case AND`.

contentgrid-automations-rest/src/main/java/com/contentgrid/automations/rest/AutomationModelVisitor.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        var path = symbolicReference.getPath();
+        try {
+            Object result = context;
+            for (var elem : path) {
+                Objects.requireNonNull(result, () -> "Cannot lookup property '%s' of null value"
+                        .formatted(getPathElementName(elem)));
+                var field = result.getClass().getDeclaredField(getPathElementName(elem));
+                field.setAccessible(true);
```
vierbergenlars, line 104: Instead of handling reflection yourself directly, I would suggest taking a look at the Spring-supplied functionality, like [`PropertyAccessor`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/PropertyAccessor.html)

contentgrid-automations-rest/src/main/java/com/contentgrid/automations/rest/AutomationsRestController.java
```diff
@@ -25,20 +27,27 @@ public class AutomationsRestController {
     private AutomationsModel model;
     @NonNull
     private final AutomationRepresentationModelAssembler assembler;
+    @NonNull
+    private final AbacContextSupplier abacContextSupplier;
+    @NonNull
+    private final AutomationModelVisitor visitor = new AutomationModelVisitor();
```
vierbergenlars, line 33:
```suggestion
    private static final AutomationModelVisitor visitor = new AutomationModelVisitor();
```

If this is final, and assigned directly: it's a constant.

contentgrid-automations-rest/src/main/java/com/contentgrid/automations/rest/AutomationsRestController.java
```diff
@@ -61,4 +70,18 @@ static AutomationsModel loadConfig(Resource resource) {
         }
     }
 
+    private List<AutomationModel> filterAutomations(List<AutomationModel> automations) {
+        var abacContext = abacContextSupplier.getAbacContext();
+        if (abacContext != null) {
```
vierbergenlars, line 75: To reduce the nesting level, you can also do 

```java
if (abacContext == null) {
    return automations;
}
```

contentgrid-spring-boot-starter/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     api platform(project(':contentgrid-spring-boot-platform'))
     api "com.contentgrid.thunx:thunx-api-spring-boot-starter"
 
+    api project(':contentgrid-automations-rest')
```
vierbergenlars, line 10:
```suggestion
    implementation project(':contentgrid-automations-rest')
```

This project does not deliver any APIs that we want to expose to the user when they add a dependency on the `contentgrid-spring-boot-starter`

vierbergenlars (changes requested):

contentgrid-automations-rest/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+dependencies {
+
+    annotationProcessor platform(project(':contentgrid-spring-boot-platform'))
+    annotationProcessor 'org.projectlombok:lombok'
+
+    implementation platform(project(':contentgrid-spring-boot-platform'))
+    implementation project(':contentgrid-spring-data-rest')
```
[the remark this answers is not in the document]
NielsCW: It is necessary for creating links to `/profile/{enitties}` and `/{entities}` (`ProfileController` and `RepositoryEntityLinks`)

contentgrid-automations-rest/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+
+    implementation 'org.springframework.data:spring-data-rest-webmvc'
+    implementation 'com.contentgrid.thunx:thunx-predicates-querydsl'
```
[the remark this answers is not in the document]
NielsCW: The dependency was still in use for `ThunkExpression`, but `thunx-predicates-querydsl` can be replaced with `thunx-model`

contentgrid-automations-rest/src/main/java/com/contentgrid/automations/rest/AutomationsRestController.java
```diff
@@ -61,4 +70,18 @@ static AutomationsModel loadConfig(Resource resource) {
...
+            return automations.stream()
+                    .filter(automation -> {
+                        var result = abacContext.accept(visitor, automation);
+                        result.assertResultType(Boolean.class);
+                        return (Boolean) result.getValue();
```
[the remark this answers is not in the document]
NielsCW: `assertResultType(Boolean.class)` gives `ThunkExpression<Boolean>` as return type which doesn't have a `getValue()` method. So this chaining only works when you add a cast `return ((Scalar<Boolean>) result.assertResultType(Boolean.class)).getValue();`

contentgrid-automations-rest/build.gradle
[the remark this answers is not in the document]
NielsCW: contentgrid-spring-data-rest is used for `ContentGridLinkRelations`

vierbergenlars (approved):

PR #248, opened 2024-07-05
Add AllowedValues constraint annotation [ACC-1466]

NielsCW:
String Attributes can now be annotated with `@AllowedValues` to limit the possible values that can be assigned to it. The options are also shown in the Hal-Forms templates

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/AllowedValues.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/AllowedValues.java
@@ -26,7 +26,7 @@ public @interface AllowedValues {
     /**
      * @return the allowed options of the constraint
      */
-    String[] options() default {};
+    String[] value();
 
     /**
      * @return the error message template
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/AllowedValues.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/AllowedValues.java
@@ -1,0 +1,46 @@
+package com.contentgrid.spring.data.rest.validation;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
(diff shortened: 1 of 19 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/AllowedValues.java
vierbergenlars, line 1: Is `contentgrid-spring-data-rest` on the compile classpath for contentgrid applications?

Because I think it should not be on the compile classpath for them, it should only be a runtime dependency (it's our internal implementation, it shouldn't have APIs to compile against.

From that follows: 
 1. contentgrid-spring-data-rest should be an `implementation` dependency of the starter
 2. validation constraints should be moved to a separate subproject that is an `api` dependency of the starter

contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/validation/AllowedValues.java
```diff
@@ -0,0 +1,61 @@
...
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * The annotated element must have one of the allowed options as value.
+ */
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(List.class)
```
vierbergenlars, line 24: It doesn't really make sense to have this constraint as `@Repeatable`.

If you have a set of allowed values, adding an additional annotation will further restrict the list of allowed values; at which point you might as well have a single annotation containing the exact list of allowed values.

Additionally, for supporting multiple annotations, you would need to evaluate the resulting restriction to generate the proper HAL-FORMS options.

NielsCW: Removed last commit

rschev (approved):

PR #256, opened 2024-07-23
Apply `@AllowedValues` constraint on all search filters

vierbergenlars:
For search filters that have a different name than the property itself,
also create the options for AllowedValues, so a select can be shown for
them as well

### DIFF
--- a/contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/hal/forms/ContentGridHalFormsConfigurationTest.java
+++ b/contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/hal/forms/ContentGridHalFormsConfigurationTest.java
@@ -92,6 +92,55 @@ class ContentGridHalFormsConfigurationTest {
     }
 
     @Test
+    void searchFormFieldForConstrainedAttributeOptions() throws Exception {
+        mockMvc.perform(get("/profile/customers").accept(MediaTypes.HAL_FORMS_JSON))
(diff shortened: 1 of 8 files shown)


contentgrid-spring-data-rest/src/test/java/com/contentgrid/spring/data/rest/hal/forms/ContentGridHalFormsConfigurationTest.java
```diff
@@ -170,9 +219,7 @@ void updateFormFieldForConstrainedAttributeOptions() throws Exception {
                                             name: "gender",
                                             type: "radio",
                                             options: {
-                                                inline: [ "female", "male" ],
-                                                minItems: 0,
-                                                maxItems: 1
```
NielsCW, line 175: Why was this removed?

rschev: The json payload for this test still contains minItems: 0, maxItems: 1, it's now just only tested in the search requests.

vierbergenlars: I think this was a rebase error; I pulled one commit out into 2 PRs. There is no reason why this shouldn't be tested.

rschev (approved):

NielsCW (approved):

PR #257, opened 2024-07-31
Add Text QuerydslPredicateFactories for starts with search [ACC-1519]

NielsCW:
This PR introduces the following text-based `QuerydslPredicateFactories` located in `Text`:
* `EqualsIgnoreCase`: case-insensitive search (existing `EqualsIgnoreCase` was moved into `Text`)
* `EqualsNormalized`: exact search utf-8 normalized with NFC
* `EqualsIgnoreCaseNormalized`: case-insensitive search utf-8 normalized with NFC
* `StartsWith`: starts with search (just added for completeness, for apps that don't use postgresql or don't have normalized indexes)
* `StartsWithIgnoreCase`: starts with case-insensitive search (just added for completeness, for apps that don't use postgresql or don't have normalized indexes)
* `StartsWithNormalized`: starts with search utf-8 normalized with NFC
* `StartsWithCaseInsensitiveNormalized`: starts with case-insensitive search utf-8 normalized with NFC


contentgrid-spring-querydsl/src/main/java/com/contentgrid/spring/querydsl/predicate/EqualsIgnoreCase.java
```diff
@@ -11,7 +11,9 @@
  * Compares strings in a case-insensitive way.
  *
  * This predicate only supports strings and can not be used with other types.
+ * @deprecated : Use {@link Text.EqualsIgnoreCase} instead.
```
vierbergenlars, line 14:
```suggestion
 * @deprecated Use {@link Text.EqualsIgnoreCase} instead.
```

contentgrid-spring-querydsl/src/main/java/com/contentgrid/spring/querydsl/predicate/EqualsIgnoreCase.java
```diff
@@ -11,7 +11,9 @@
...
  */
+@Deprecated(since = "v0.15.2", forRemoval = true)
 public class EqualsIgnoreCase extends AbstractSimpleQuerydslPredicateFactory<StringPath, String> {
```
vierbergenlars, line 17: I think you can make this one extend `Text.EqualsIgnoreCase` instead? 
Keeping the class for backwards compatibility, but already delegating the implementation to the new version.

vierbergenlars (approved):

PR #260, opened 2024-08-13
ACC-1531 Add failing test that receives 500 instead of 400

rschev:


rschev:
Test will pass with the next spring-content release (3.0.14).

vierbergenlars (approved):

PR #270, opened 2024-09-05
Only handle jackson exceptions when wrapped inside HttpMessageNotReadableException

vierbergenlars:
Handling Jackson exceptions generically means that they can also be triggered by other parts of the request-response cycle,
including during response serialization.
Sending a 400 error for things that are not related to actual input message parsing is bad form
and makes it more difficult to diagnose the actual issue (compared to having a HTTP 500 happen for an unhandled exception)

Due to the setup for `@ExceptionHandler` methods from Spring, all cases for an exception need to be handled in one method.
There is no way to automatically dispatch different methods based on the nested cause of an exception.

### DIFF
--- a/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
+++ b/contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
@@ -20,12 +20,11 @@ import org.hibernate.exception.ConstraintViolationException;
 import org.springframework.context.support.MessageSourceAccessor;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.rest.core.RepositoryConstraintViolationException;
-import org.springframework.data.rest.core.config.RepositoryRestConfiguration;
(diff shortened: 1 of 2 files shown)


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/rest/problem/ContentGridExceptionHandler.java
```diff
@@ -101,6 +100,26 @@ ResponseEntity<Problem> handleRepositoryConstraintViolationException(RepositoryC
...
+                return handleJsonParseException(parseException);
+            }
+            currentException = currentException.getCause();
+        }
+
+        // Fallback handler: just a generic bad request
+        return responseEntityFactory.createResponse(
+                problemFactory.createProblem(ProblemType.INVALID_REQUEST_BODY)
+                        .withStatus(HttpStatus.BAD_REQUEST)
+                        .withDetail(exception.getMessage())
```
rschev, line 119: Sonar complains about `exception` being null here, but it's not `currentException`, which could actually be null.

rschev (approved):

PR #272, opened 2024-09-10
Pagination for large datasets: Use estimated counts for total number of items in a collection

vierbergenlars:
1. Install a replacement `QuerydslJpaPredicateExecutor` that support custom count strategies when paginating queries
2. Implement count strategies
  - Plain `count(*)` with an execution timeout of 1 second
  - Estimated count using `EXPLAIN <query>` and extracting the estimated RowCount from the postgres query plan

Counts are packed up in a `CountResult`, which is placed in `ApproximatelyCountedPage` (a custom implementation of `Page`).

The quality of the count (exact or estimated) is not part of the REST layer yet, currently all counts are shown as `totalElements`/`totalPages`, even when they are estimated.
REST layer functionality is for a separate PR.


contentgrid-spring-data-counting/src/test/java/com/contentgrid/spring/data/counting/ApproximatelyCountedPageTest.java
```diff
@@ -0,0 +1,188 @@
...
+            assertThat(page.getCountResult()).isEqualTo(CountResult.exact(20));
+            assertThat(page.hasNext()).isFalse();
+            assertThat(page.hasPrevious()).isTrue();
+        }
+
+        @Test
+        void hasNextPage_countJustTooSmall() {
+            var page = createPage(PageRequest.ofSize(10), 11, CountResult.estimated(10));
+
+            // There is no next page, so we know that this page has the final number of results
```
NielsCW, line 101: comment is not correct

contentgrid-spring-data-counting/src/main/java/com/contentgrid/spring/data/counting/ApproximatelyCountedPage.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    public Pageable nextOrLastPageable() {
+        return super.nextOrLastPageable();
+    }
+
+    @Override
+    public Pageable previousOrFirstPageable() {
+        return super.previousOrFirstPageable();
+    }
```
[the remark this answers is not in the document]
vierbergenlars: <s>
I was convinced that they are necessary because of the multiple inheritance diamond of death.

`ApproximatelyCountedPage` implements the `Slice` interface via 2 paths, and both paths have an implementation of `nextOrLastPageable()` (the default implementation in `Slice`).

```mermaid
classDiagram
    Slice <|-- SliceImpl
    Slice <|-- Page
    Page <|-- ApproximatelyCountedPage
    SliceImpl <|-- ApproximatelyCountedPage

    class Slice {
        +nextOrLastPageable()
    }
```

The actual overrides may change: `Page` or `SliceImpl` (or both) may provide their own default implementation, and as a subclass, we need to pick which one we want to invoke.
</s>

Looks like I can remove the methods, and it still compiles. I am quite surprised by that, because I was certain I added the overrides only because it wouldn't compile otherwise.

contentgrid-spring-data-pagination/src/main/java/com/contentgrid/spring/data/pagination/jpa/hibernate/CountExplainPostgreSQLDialect.java
thijslemmens, line 1: I think we can only use one dialect at a time. If we need to extend the dialect further for another reason, we'll have to make a CGPostgreSQLDialect.

contentgrid-spring-data-pagination/src/test/java/com/contentgrid/spring/data/pagination/jpa/strategy/TimedDirectCountItemCountStrategyTest.java
```diff
@@ -0,0 +1,125 @@
...
+                        assertThat(result.count()).isEqualTo(25);
+                    });
+        } finally {
+            transactionManager.commit(tx);
+        }
+    }
+
+    @Test
+    void performCount_outOfTime() throws SQLException {
+        try (var conn = dataSource.getConnection()) {
```
thijslemmens, line 96: This might need a bit of explanation

contentgrid-spring-data-pagination/src/main/java/com/contentgrid/spring/data/pagination/jpa/hibernate/CountExplainPostgreSQLDialect.java
vierbergenlars: It could be implemented as a dialect wrapper instead, but then I needed to override 3 or 4 methods that delegated to `getQueryHintString`; which I did not really think made more sense.

contentgrid-spring-data-pagination/src/test/java/com/contentgrid/spring/data/pagination/jpa/strategy/TimedDirectCountItemCountStrategyTest.java
vierbergenlars: This line, or the view below? And how much do I have explain? 

- The search path from the init-sql; how this view gets put inbetween
- what a `cross join lateral` is
- how the sleep works

The part that I think is obvious is that it makes the query slower.
Is it necessary to explain all the details of how it works for some test code?

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/data/pagination/JpaPaginationAutoConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+public class JpaPaginationAutoConfiguration {
+
+    // define this bean just so the built-in JpaRepositoriesAutoConfiguration does not trigger.
+    // otherwise all repositories will be registered *twice*, which will cause conflicts
+    @Bean
+    JpaRepositoryConfigExtension jpaRepositoryConfigExtension() {
+        return new JpaRepositoryConfigExtension();
+    }
+
+    static class JpaRepositoriesImportSelector implements ImportSelector {
```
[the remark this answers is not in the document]
vierbergenlars: I don't understand how this works exactly either; I copied and adapted it from the spring `JpaRepositoriesAutoConfiguration` and tried to put as much comments about the things that  I understood myself in here.

thijslemmens (approved):

PR #275, opened 2024-09-18
Support cursor-based pagination

vierbergenlars:
Initial support for cursor-based pagination.

This PR contains mostly API changes to support changing to an actual cursor-based pagination at a later point, without requiring API changes at that point.

1. Page metadata gets `prev_cursor` and `next_cursor` properties (for the benefit of users not using HAL)
2. Enforcement of proper cursor-based usage of the API instead of a page parameter that an be controlled client-side
   - query parameter is named `_cursor` by default
   - cursor-based pagination can be enabled via `contentgrid.rest.pagination=page_cursor` configuration property
   - HAL `prev`, `next`, `first` and `self` links are automatically using the correct parameters
3. Pagination parameter errors result in HTTP 400 Bad Request instead of silent fallbacks to the default values
   - Silent fallback to defaults make it hard for the API user to know why their request was not honored, and can result in accidental incorrect pagination
   - The only exception is page size: When a too large page size is supplied, it is silently clamped to the maximum size. This is done because the maximum page size is not easily discoverable, and the maximum size is not the default size.

To maintain backwards compatibility, the default `contentgrid.rest.pagination=page_number` keeps using the `page` query parameter with a plain page number.
`prev_cursor` and `next_cursor` can be still be used in that case, but their usage is not mandatory


contentgrid-spring-data-pagination/src/main/java/com/contentgrid/spring/data/pagination/web/ContentGridSpringDataPaginationWebConfiguration.java
```diff
@@ -11,15 +13,20 @@
...
+            ObjectProvider<CursorEncoder> cursorEncoder
     ) {
         return new BeanPostProcessor() {
             @Override
             public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
                 if (bean instanceof PagedResourcesAssembler<?>) {
-                    return new ItemCountPageResourceAssembler<>(pageableHandlerMethodArgumentResolver);
+                    return new ItemCountPageResourceAssembler<>(
+                            pageableHandlerMethodArgumentResolver,
+                            org.springframework.data.util.Lazy.of(cursorEncoder::getIfUnique)
```
thijslemmens, line 28: Won't this possibly result in NullPointers when there is no CursorEncoder or when there are multiple?

thijslemmens (changes requested):

contentgrid-spring-data-pagination/src/main/java/com/contentgrid/spring/data/pagination/web/ContentGridSpringDataPaginationWebConfiguration.java
vierbergenlars: There is a check at the usage site, falling back to an empty `CursorPageMetadata` if there is no `CursorEncoder` in the `Lazy`, so no null pointer exceptions.

https://github.com/xenit-eu/contentgrid-spring/blob/927e93c6c7f468bf2188ebee39b61a3df438be6b/contentgrid-spring-data-pagination/src/main/java/com/contentgrid/spring/data/pagination/web/ItemCountPageResourceAssembler.java#L74-L79

The result of a `CursorPageMetadata` with null values is that no `prev_cursor` or `next_cursor` fields get rendered.

You did find an issue here though: if there are multiple `CursorEncoder`s (which *should not happen*), there won't be any cursors rendered either. I think we prefer an actual exception in that case indicating that there is an error.

Unfortunately, I don't think there is a way to make the error appear during startup, we don't want to have any bean eagerly-initialized for injection in a `BeanPostProcessor`.

So I think we should replace the `getIfUnique` with `getIfAvailable` (which will throw when there are duplicate beans without any primary or priority difference)

```suggestion
                            org.springframework.data.util.Lazy.of(cursorEncoder::getIfAvailable)
```

thijslemmens (approved):

PR #282, opened 2024-10-11
ACC-1662: Omit sort property when there are no sort-options present

NielsCW:


contentgrid-spring-data-rest/src/main/java/com/contentgrid/spring/data/querydsl/sort/CollectionFilterSortHalFormsPayloadMetadataContributor.java
```diff
@@ -63,8 +64,11 @@ public Stream<PropertyMetadata> contributeToSearchForm(Class<?> domainType) {
                             messageResolver.resolve(new DirectionMessageSourceResolvable(domainType, filter, dir)),
                             foldIntoExpressions(Sort.by(dir, filter.getFilterName()))
                     ));
-                });
-        return Stream.of(new SortPropertyMetadata(sortOptions.toList()));
+                }).toList();
+
+        return Stream.of(sortOptions)
+                .filter(Predicate.not(List::isEmpty))
+                .map(SortPropertyMetadata::new);
```
vierbergenlars, line 71: Not everything needs to be done with streams.

Writing an if instead of doing stream manipulation makes the logic easier to follow

```suggestion
        if (sortOptions.isEmpty()) {
            return Stream.empty();
        } else {
            return Stream.of(new SortPropertyMetadata(sortOptions));
        }
```

vierbergenlars (approved):

PR #291, opened 2024-11-18
Ensure that HalConfiguration MediaTypeConfigurationCustomizers are always run

vierbergenlars:
`cg:content` & `cg:relation` links are not an array when there is only a single value.

This happens because spring-data-rest does not process `MediaTypeConfigurationCustomizer`, so it does not apply the configuration when pulling the `HalConfiguration` bean directly.
This leads to the configuration settings in `HalConfiguration` inconsistently being applied, depending on if a setting is used in spring-hateoas or spring-data-rest.

Always applying all customizers directly in a `BeanPostProcessor` ensures that the HalConfiguration bean itself has all the customizations directly baked-in,
and does not require all consumers to be aware of `MediaTypeConfigurationCustomizer`.

This _does_ result in `MediaTypeConfigurationCustomizer`s being applied twice where `HalConfiguration` is used in spring-hateoas,
but all usual configurations are idempotent anyways (e.g. adding configuration for a link relation, or setting the ObjectMapper customizer)


NielsCW (commented):
There seems to be a failing test in `AutomationsRestControllerTest`.

vierbergenlars:
I noticed.
We need to see how we are going to tackle that, because the `cg:entity` relation can only be configured *once* to be either be single or a list.
The main endpoints really need it to be a list; the automation endpoint seems to use it as a pointer to a single item. Those usages are not compatible.

NielsCW (approved):

PR #293, opened 2024-11-22
Move automations linkrel types to separate namespace

vierbergenlars:
The automation annotation endpoint should have their own namespace for link relation types.

For one, the automation annotations currently reuse the `cg:entity` linkrel to reference the entity collection. However, this causes a problem when fixing #291, because `cg:entity` is defined to be multi-valued; while this relation is single-valued.

Since we need a couple of separate link-relation type anyways, let’s reserve the automation CURIE prefix (and a separate URL namespace) to use for link relation types related to automations. This clearly separates the concepts from contentgrid data model concepts.

### DIFF
--- a/contentgrid-automations-rest/src/main/java/com/contentgrid/spring/automations/AutomationLinkRelations.java
+++ b/contentgrid-automations-rest/src/main/java/com/contentgrid/spring/automations/AutomationLinkRelations.java
@@ -14,5 +14,6 @@ public class AutomationLinkRelations {
     public static final LinkRelation ANNOTATION = HalLinkRelation.curied(CURIE, "annotation");
     public static final String ANNOTATION_STRING = CURIE+":annotation";
     public static final LinkRelation TARGET_ENTITY = HalLinkRelation.curied(CURIE, "target-entity");
+    public static final LinkRelation REGISTRATIONS = HalLinkRelation.curied(CURIE, "registrations");
 
 }
--- a/contentgrid-automations-rest/src/main/java/com/contentgrid/spring/automations/AutomationLinkRelations.java
+++ b/contentgrid-automations-rest/src/main/java/com/contentgrid/spring/automations/AutomationLinkRelations.java
@@ -1,4 +1,4 @@
-package com.contentgrid.automations;
+package com.contentgrid.spring.automations;
 
 
 import lombok.experimental.UtilityClass;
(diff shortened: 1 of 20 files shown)


vierbergenlars:
Since the second commit is just moving a bunch of files to a different package, you probably want to review this PR commit-by-commit.

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/automation/ContentGridAutomationAutoConfiguration.java
```diff
@@ -1,35 +1,18 @@
...
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
 
-@AutoConfiguration
+@AutoConfiguration(
+        after = ContentGridSpringDataRestAutoConfiguration.class
+)
 @ConditionalOnWebApplication
-@ConditionalOnClass({ AutomationsRestController.class, AbacContextSupplier.class, ThunkExpression.class })
```
NielsCW, line 18: Are `AbacContextSupplier` and `ThunkExpression` automatically on the classpath when `ContentGridAutomationsConfiguration` is on the classpath?

contentgrid-automations-rest/src/main/java/com/contentgrid/spring/automations/AutomationLinkRelations.java
```diff
@@ -14,5 +14,6 @@ public class AutomationLinkRelations {
     public static final LinkRelation ANNOTATION = HalLinkRelation.curied(CURIE, "annotation");
     public static final String ANNOTATION_STRING = CURIE+":annotation";
     public static final LinkRelation TARGET_ENTITY = HalLinkRelation.curied(CURIE, "target-entity");
+    public static final LinkRelation REGISTRATIONS = HalLinkRelation.curied(CURIE, "registrations");
```
rschev, line 17: Wouldn't this just be `automations:automations`? Because it's not like you can register at that endpoint.

contentgrid-spring-boot-autoconfigure/src/main/java/com/contentgrid/spring/boot/autoconfigure/automation/ContentGridAutomationAutoConfiguration.java
vierbergenlars: I might have been a bit too optimistic about that

contentgrid-automations-rest/src/main/java/com/contentgrid/spring/automations/AutomationLinkRelations.java
thijslemmens: It's right that you cannot add registrations here, but you can list the registrations that already there.

thijslemmens (approved):

NielsCW (approved):

PR #314, opened 2025-02-13
Add contentgrid-spring-content-encryption for encrypted content [ACC-1848]

NielsCW:
Currently only adding a `DataEncryptionKeyAccessor` that stores the unencrypted keys in a separate table.


contentgrid-spring-boot-starter/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     api project(':contentgrid-spring-data-support')
     api project(':contentgrid-spring-querydsl')
     api project(':contentgrid-spring-common')
+    api project(':contentgrid-spring-content-encryption')
```
thijslemmens, line 16: Shouldn't this be `impementation`?

contentgrid-spring-content-encryption/src/main/java/com/contentgrid/spring/content/encryption/ContentDataEncryptionKey.java
```diff
@@ -0,0 +1,50 @@
...
+
+@Entity
+@Table(name = TABLE_NAME)
+@Getter
+@Setter
+@NoArgsConstructor
+public class ContentDataEncryptionKey {
+    static final String TABLE_NAME = "_dek_storage";
+
+    public static final org.jooq.Table<Record> DEK_STORAGE = table(name(TABLE_NAME));
```
rschev, line 26: Did you check whether this respects the `spring.datasource.hikari.schema` property?

NielsCW: Tried it out and if there is a view on `"V1"."_dek_storage"` and `spring.datasource.hikari.schema=V1`. It is able to store and load deks from `_dek_storage`. Even if there is a `"V2"` that renames the columns of `_dek_storage` (using  `V2.0__rewrite_V1_for_next.sql`, `V2.2__rewrite_V2_to_current.sql` and `V2.3__rewrite_V1_to_V2.sql`).

thijslemmens (approved):

contentgrid-spring-content-encryption/src/main/java/com/contentgrid/spring/content/encryption/TableStorageDataEncryptionKeyAccessor.java
```diff
@@ -0,0 +1,71 @@
...
+import lombok.RequiredArgsConstructor;
+import org.jooq.DSLContext;
+import org.springframework.content.commons.mappingcontext.ContentProperty;
+import org.springframework.content.encryption.keys.DataEncryptionKeyAccessor;
+import org.springframework.content.encryption.keys.StoredDataEncryptionKey.UnencryptedSymmetricDataEncryptionKey;
+
+@RequiredArgsConstructor
+public class TableStorageDataEncryptionKeyAccessor<S> implements DataEncryptionKeyAccessor<S, UnencryptedSymmetricDataEncryptionKey> {
+    private final DSLContext dslContext;
+    private final String wrappingKeyLabel;
```
vierbergenlars, line 20: Why do you have to specify a wrapping key label? There is no key that wraps the DEK, so it should be `NULL` in the database?

vierbergenlars (changes requested):

contentgrid-spring-content-encryption/src/main/java/com/contentgrid/spring/content/encryption/TableStorageDataEncryptionKeyAccessor.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            dslContext.insertInto(DEK_STORAGE)
+                    .set(CONTENT_ID, contentId)
+                    .set(ALGORITHM, dataEncryptionKey.getAlgorithm())
+                    .set(ENCRYPTED_DEK, dataEncryptionKey.getKeyData())
+                    .set(INITIALIZATION_VECTOR, dataEncryptionKey.getInitializationVector())
+                    .set(KEK_LABEL, wrappingKeyLabel)
+                    .execute();
+        }
```
vierbergenlars, line 67: Is this running inside a transaction?
Otherwise there is a risk of deleting the keys without the new keys being inserted at all if the process crashes in-between

contentgrid-spring-content-encryption/src/main/java/com/contentgrid/spring/content/encryption/TableStorageDataEncryptionKeyAccessor.java
NielsCW: `kek_label` cannot be nullable because it is part of the primary key. Should only the content_id be part of the primary key?

rschev: No, because we need to be able to have multiple encrypted DEKs per content id when we do implement KEKs.

rschev (approved):

vierbergenlars (approved):

PR #323, opened 2025-02-24
Add extension registration properties

vierbergenlars:
Extension registration is used for supplying extension base URIs that are dependent on the deployment environment.
This way, it can be avoided to hardcode domains for external services, which might change independently of the application itself.

### DIFF
--- a/contentgrid-spring-common/src/main/java/com/contentgrid/spring/common/ContentGridExtensionProperties.java
+++ b/contentgrid-spring-common/src/main/java/com/contentgrid/spring/common/ContentGridExtensionProperties.java
@@ -1,0 +1,25 @@
+package com.contentgrid.spring.common;
+
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import lombok.Data;
+
+@Data
+public class ContentGridExtensionProperties {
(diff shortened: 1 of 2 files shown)


NielsCW (commented):
Should these properties only be used by the (generated) application code?

vierbergenlars:
At them moment, they are only used by generated application code indeed, but I left enough room to add additional configuration settings that can potentially be used by other code as well.

contentgrid-spring-common/src/main/java/com/contentgrid/spring/common/ContentGridExtensionProperties.java
```diff
@@ -0,0 +1,25 @@
...
+public class ContentGridExtensionProperties {
+    private Map<String, ExtensionRegistration> registration = new HashMap<>();
+
+    public Optional<ExtensionRegistration> getRegistration(String extensionName) {
+        return Optional.ofNullable(registration.get(extensionName));
+    }
+
+    @Data
+    public static class ExtensionRegistration {
+        private Map<String, URI> basePath = new HashMap<>();
```
thijslemmens, line 19: Can you add some examples of how different runtime platforms map to these ExtensionRegistrations?

NielsCW (approved):

